PR #110, opened 2026-08-27
literature: distinguish transport failures from genuine misses in resolve

### LINKED ISSUE #106: literature: graph.resolve cannot distinguish a transport failure from a genuine miss
## Context

This project's failure-honesty rule (CLAUDE.md: "never let a failure or
uncertain condition be silently reported as a legitimate empty/negative/
complete result... distinguish 'failed' from 'legitimately empty'") applies
directly to `graph.resolve`
(`defendable-science/defendable_science/literature/graph.py:182`), and today
it doesn't hold.

`resolve` re-raises `RateLimitError` (so a throttle correctly propagates
rather than being recorded as a miss), but folds every other `HttpError` into
the same shape a genuine "no such paper" result gets:

```python
try:
    work = client.get_json(lookup)
except RateLimitError:
    raise
except HttpError as exc:
    return {"resolved": False, "reason": str(exc)}
```

(`graph.py:208-211`). A consumer of `literature resolve`'s JSON output sees
only `{"resolved": false, "reason": "..."}` in both cases — a transport
failure (a 502, a DNS hiccup, a timeout wrapped as `HttpError`) and a genuine
miss (404 / empty body / no cross-reference) are indistinguishable from the
outside.

Inside `acquire_one` (`defendable-science/defendable_science/literature/
acquire.py:1640`), the distinction happens to survive by a different path: a
resolution failure there becomes an `errors[]` row (`BUCKET_ERROR`), never
`manual[]`, per the docstring at `acquire.py:1909` ("registry becomes an
`errors[]` row rather than a crash"). But that's specific to `acquire_one`'s
own error handling around `_resolve_work`, not a property of `resolve`
itself — a direct caller of `defendable-science literature resolve <id>` gets
no such signal.

This is pre-existing (predates the literature-asset-acquisition branch) and
was deliberately not changed during that work, because fixing it would alter
shipped CLI output for a command (`resolve`) the branch wasn't otherwise
touching.

## Goal

A consumer of `literature resolve`'s JSON can tell a transport failure apart
from a genuine miss, and the CLI's exit code reflects the distinction.

## Where

- `defendable-science/defendable_science/literature/graph.py:182`
  (`resolve`) — the function; the fold-together happens at lines 208-211
  (`except HttpError as exc: return {"resolved": False, "reason": str(exc)}`).
- `defendable-science/defendable_science/literature/acquire.py:1909`
  (docstring note) and the `_resolve_work` call inside `acquire_one`
  (`acquire.py:1640`) — where the distinction currently survives by a
  different, narrower path.
- `defendable-science/defendable_science/cli.py:275` (`literature.resolve`
  command) — where the CLI's exit code is set from `resolve`'s return value
  and would need to change to reflect the new discriminator.

## Proposed approach

Add a discriminator field to the failure shape — e.g. `transport_error:
true`, or replace the single `resolved: bool` with a tri-state `status`
(`"resolved" | "not_found" | "transport_error"`), keeping `resolved` for
backward compatibility if a tri-state is judged too disruptive. Populate it
at the `except HttpError as exc:` branch (`graph.py:211`) vs. the existing
`{"resolved": False, "reason": "no work found"}` / `"unsupported identifier
kind"` / cross-reference-failure branches, which are genuine misses and
should NOT set the new flag.

Then thread the CLI's exit code (`cli.py:275`'s `resolve` command) so a
transport error exits differently from a clean miss — e.g. exit 2 for
transport failure vs. exit 1 (or 0, if a miss is treated as a legitimate
non-error result) for a genuine miss, matching whatever convention the rest
of the `literature`/`dataset` CLI already uses for degraded-vs-clean results.

## Acceptance criteria

- [ ] `resolve`'s return value distinguishes a transport failure
      (`HttpError` other than `RateLimitError`) from a genuine miss (404 /
      empty body / unresolvable cross-reference / unsupported identifier
      kind), via an explicit field or status value — not just `reason`'s free
      text.
- [ ] `RateLimitError` still propagates unchanged (no regression on that
      behaviour).
- [ ] The `literature resolve` CLI command's exit code reflects the
      distinction.
- [ ] `acquire_one`'s existing `errors[]`-vs-`manual[]` behaviour is
      unaffected (still correct after the change).
- [ ] Tests cover: a genuine 404 miss, a transport error (e.g. 502 or a
      wrapped connection failure), and a rate limit, asserting each produces
      a distinguishable result; 100% branch coverage held (ADR-0028).

## References

- `defendable-science/defendable_science/literature/graph.py:182-221`
  (`resolve`)
- `defendable-science/defendable_science/literature/acquire.py:1640`
  (`acquire_one`), `:1909` (docstring)
- `defendable-science/defendable_science/cli.py:275` (`literature.resolve`
  command)
- CLAUDE.md — failure-honesty rule
- Related: #97 Gap 1 (the acquisition work this was found alongside, but
  deliberately not fixed in)

davorrunje:
## Summary

`graph.resolve()` re-raised `RateLimitError` but folded every other
`HttpError` — a `502`, an exhausted retry budget, a non-JSON body — into
the same `{resolved: False, reason: ...}` shape a genuine "no such paper"
result gets. A consumer of `literature resolve`'s JSON output couldn't
tell a transport failure from a genuine miss, violating this repo's
failure-honesty rule (CLAUDE.md).

## Design (additive, not a signature change)

- `HttpError` (`defendable_science/core/http.py`) now carries an optional
  `status_code: int | None`, set at the one place a single non-retryable
  status is known (`HTTP {status}` raise). It stays `None` when there's no
  single status to point to — retries exhausted on a shifting mix of
  `5xx`/`429`, a network exception, or a non-JSON `200` body.
- `resolve()` (`literature/graph.py`) now checks `exc.status_code == 404`
  in its `except HttpError` branch: a genuine 404 keeps today's shape;
  anything else additionally sets `transport_error: True`. I chose an
  **additive flag over a tri-state `status`** because `resolve` is a
  shipped command other things may already parse on `resolved`/`reason` —
  adding a field can't break an existing consumer that ignores unknown
  keys, whereas replacing `resolved: bool` could.
- Per the issue's explicit guidance, the S2 cross-reference-miss branch and
  the "unsupported identifier kind" / "no work found" (empty body)
  branches are **not** touched — those remain genuine misses even though
  `_s2_crossref` folds its own `HttpError`s the same way today. That's
  called out as intentional scope, not an oversight.
- `literature resolve`'s CLI exit code now follows the distinction: `0`
  resolved, `1` genuine miss, `2` transport failure. (Rate limits keep
  exiting `1` via the pre-existing `_http_guard`, unrelated to this fix.)
- `acquire_one`'s `_resolve_work` was checked and needs **no change**: it
  already turns *any* `not info.get("resolved")` into an `errors[]` row
  regardless of `reason` text, so it was never inferring from the string
  and is unaffected by the new field.

## Test plan

- [x] New test: a 502 that exhausts the retry budget sets
      `transport_error: True` at the `graph.resolve` layer and exits `2`
      via the CLI, and is asserted distinguishable from a 404 miss.
- [x] New test: a non-JSON `200` body is also `transport_error: True`
      (not a miss).
- [x] Negative assertions added to the existing 404/unsupported-kind/
      empty-body/S2-crossref-miss tests: `transport_error` must be absent.
- [x] `RateLimitError` propagation test (`test_resolve_rate_limit_propagates_not_miss`)
      unchanged/still passing.
- [x] `HttpError.status_code` asserted at the `http.py` layer for both the
      immediate-4xx raise (`404`) and the exhausted-retries raise (`None`).

Gate, run for real from `defendable-science/`:

```
$ uv run ruff check . && uv run ruff format --check . && uv run mypy && uv run pytest -q
All checks passed!
39 files already formatted
Success: no issues found in 41 source files
639 passed, 14 skipped in 4.67s
TOTAL coverage: 100.00%
```

From the worktree root:

```
$ ./tools/validate-plugin.sh
✔ Validation passed

$ uvx pre-commit run --all-files
...all hooks Passed (trim whitespace, end-of-file, yaml, large-files,
pyupgrade, codespell, ruff lint+format, mypy, plugin-validate, detect-secrets)
```

Closes #106

### DIFF
--- a/defendable-science/defendable_science/cli.py
+++ b/defendable-science/defendable_science/cli.py
@@ -276,12 +276,22 @@ def _openalex_id(client: HttpClient, identifier: str) -> str:
 def resolve(identifier: str) -> None:
     """Resolve an identifier (DOI, arXiv id, OpenAlex/S2 id) to a canonical work.
 
+    The exit code follows the result so a caller need not parse JSON to tell
+    the three outcomes apart: ``0`` on a resolution, ``1`` on a genuine miss
+    (no such paper), ``2`` on a transport failure (``transport_error: true``
+    in the JSON) — never reported as a clean "no result".
+
     :param identifier: The identifier to resolve.
     """
     client = _lit_client()
     with _http_guard(client):
-        typer.echo(json.dumps(graph_mod.resolve(identifier, client=client), indent=2))
-        raise typer.Exit(code=0)
+        record = graph_mod.resolve(identifier, client=client)
+        typer.echo(json.dumps(record, indent=2))
+        if record.get("resolved"):
+            raise typer.Exit(code=0)
+        if record.get("transport_error"):
+            raise typer.Exit(code=2)
+        raise typer.Exit(code=1)
 
 
 @literature.command()
--- a/defendable-science/defendable_science/core/http.py
+++ b/defendable-science/defendable_science/core/http.py
@@ -32,7 +32,21 @@ _T = TypeVar("_T")
 
 
 class HttpError(RuntimeError):
-    """Raised when a request ultimately fails (after retries) or returns non-JSON."""
+    """Raised when a request ultimately fails (after retries) or returns non-JSON.
+
+    :param message: The error message.
+    :param status_code: The response's HTTP status, when the failure is a
+        known non-retryable status (e.g. ``404``). ``None`` when the failure
+        has no single status to point to — retries exhausted on a shifting
+        mix of ``5xx``/``429``, a network exception, or a non-JSON ``200``
+        body. Callers use this to tell a genuine miss (``404``) apart from a
+        transport failure, per the failure-honesty rule.
+    """
+
+    def __init__(self, message: str, *, status_code: int | None = None) -> None:
+        """Build the error; see the class docstring for parameter semantics."""
+        super().__init__(message)
+        self.status_code = status_code
 
 
 class RateLimitError(HttpError):
@@ -335,7 +349,10 @@ class HttpClient:
                 if resp.status_code == 200:
                     return extract(resp)
                 if resp.status_code not in (429, 500, 502, 503, 504):
-                    raise HttpError(f"{url}: HTTP {resp.status_code}")
+                    raise HttpError(
+                        f"{url}: HTTP {resp.status_code}",
+                        status_code=resp.status_code,
+                    )
                 last_error = f"HTTP {resp.status_code}"
                 retry_after = _retry_after_seconds(resp.headers)
                 rate_limited = resp.status_code == 429 or (
--- a/defendable-science/defendable_science/literature/graph.py
+++ b/defendable-science/defendable_science/literature/graph.py
@@ -185,8 +185,11 @@ def resolve(identifier: str, *, client: HttpClient) -> dict[str, Any]:
     :param identifier: DOI / arXiv id / OpenAlex ``W…`` / S2 id.
     :param client: The HTTP client.
     :returns: ``{resolved, openalex, doi, s2, arxiv, title, year}``; on a genuine
-        miss (``404`` / empty body / no cross-reference), ``{resolved: False,
-        reason: …}``.
+        miss (``404`` / empty body / no cross-reference / unsupported identifier
+        kind), ``{resolved: False, reason: …}``. On a transport failure (any
+        other ``HttpError`` — a ``502``, an exhausted retry budget, a non-JSON
+        body), the same shape additionally carries ``transport_error: True`` —
+        a consumer must not treat that as "no such paper".
     :raises RateLimitError: If a provider rate-limits — a throttle must propagate,
         never be recorded as a "not found".
     """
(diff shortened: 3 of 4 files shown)


davorrunje:
Follow-up fix per review: exit code 2 was already Click/Typer's usage-error code (and used elsewhere in the `literature` group for argument misuse in `fetch`/`confirm`/`verify`/`mirror`), so mapping the transport failure to 2 collided within `resolve` itself — a bad flag and a network failure both exited 2.

Moved the transport failure to **exit 3**. Final mapping for `literature resolve`, documented in its `:raises typer.Exit:` docstring:

- `0` — resolved
- `1` — genuine miss (no such paper)
- `2` — Click/Typer usage error (unchanged, not raised by this command's body)
- `3` — transport failure (`transport_error: true` in the JSON)

Added `test_cli_resolve_transport_error_exits_3` and, as the negative assertion, `test_cli_resolve_usage_error_still_exits_2_not_3` (missing-argument invocation still exits 2 and never emits `transport_error`).

Checked for any other place documenting `literature resolve`'s exit codes — `docs/guides/literature.md`, `skills/literature/SKILL.md`, `docs/USER-GUIDE.md`, and the design docs — none exists, so nothing else needed updating.

Gate, re-run for real from `defendable-science/`:

```
$ uv run ruff check . && uv run ruff format --check . && uv run mypy && uv run pytest -q
All checks passed!
39 files already formatted
Success: no issues found in 41 source files
640 passed, 14 skipped in 3.40s
TOTAL coverage: 100.00%
```

From the worktree root:

```
$ ./tools/validate-plugin.sh
✔ Validation passed

$ uvx pre-commit run --all-files
...all hooks Passed
```

New commit: 471ef3d

PR #175, opened 2026-08-28
test(package): guard every documented CLI invocation against the Typer tree

davorrunje:
## What

Adds `defendable-science/tests/test_skill_cli_contract.py`, which extracts every
documented `defendable-science …` / `dsci …` invocation from `skills/**`,
`resources/**`, `docs/guides/`, `docs/USER-GUIDE.md` and `README.md` and resolves
each against the live Typer tree — commands, subcommands, flags and argument
arity. Four invocations were shipping broken; all four are fixed here, and each
was reproduced by the guard before being fixed.

This is API-2 and API-3 from the audit (`.codebase_audit/api-tech-debt.md`),
which rated API-3 *"the highest-value missing test in the repository"*.

## The four breaks

| Where | Problem |
|---|---|
| `docs/USER-GUIDE.md:182` | `--EIG` — the option is `--eig`; `EIG` is the backlog table *column*. Exit 2. |
| `docs/guides/dataset.md:32` | `dataset verify` with no identifier — in the block that establishes the guide's own "a shell block is a real command, verbatim" convention. |
| `docs/guides/dataset.md:139` | Same missing identifier. |
| `skills/hypothesis-exploration/SKILL.md:56` | Header documented as `one-line hypothesis`; the column is `one-line`. |

The last one is the load-bearing one: `skills/hypothesis-exploration/SKILL.md:79`
tells the agent to keep that column order *"so the `backlog` verbs can parse
it"*, and `Backlog._append` refuses a table built to the documented spec. A
second, tiny test pins the documented header against `columns_for("hypothesis")`
so prose and profile cannot drift again.

## Details

- **Why nothing caught this.** `tests/test_plugin_content.py` is the only test
  that opens a `SKILL.md` and it *deliberately skips fenced blocks*, which is
  where every invocation lives. `tests/test_cli_help.py` walks the Click tree but
  only checks help text for MyST leakage. `.pre-commit-config.yaml:69` scopes the
  plugin hook to `^\.claude-plugin/.*\.json$`. The guard closes that gap from the
  other side; it needs no CI wiring, since it is an ordinary pytest file.

- **Structure, not values.** Option values are never type-converted, because
  documentation writes placeholders (`<id>`, `<paper>`) where a real run takes
  data. Checking a placeholder against a param's type would be noise.

- **A fence is a command; an inline span is a reference.** `docs/guides/dataset.md`
  states the convention itself, against quote blocks the reader is "not meant to
  type into a shell". So a fenced line is held to full argument arity, while an
  inline span is checked only for the names it uses. This is what lets
  `resources/templates/README.md:53` list `backlog promote --scaffold` in a
  "what writes this file" table without a row id.

- **Four false-positive classes**, each found by running the guard against the
  real corpus and each with its own regression test: trailing shell comments
  counted as positional arguments; usage-synopsis grammar (`[--size N]`,
  `(--citekey KEY ...| --all)`); the `< keys.json` input redirect; and prose that
  documents a command's *absence* — "There is no `defendable-science literature
  scout` command" appears in three guides and must not read as drift.

- **Deliberately not covered.** `docs/design/` and `docs/superpowers/` are
  excluded as dated design records and historical plans, where a flag that never
  shipped is an accurate record. The audit's API-12 counts nine such invocations
  there; including them would mean either editing history or nine exemptions.

- **`dataset.md:139` — a judgement call worth a second opinion.** The audit
  suggested switching this to `dataset audit`, since the surrounding prose asks a
  whole-manifest question. I used `verify imagenet-c` instead: the following
  sentence is explicitly about `verify`'s offline guarantee ("`verify` never
  downloads. That is the point"), and `audit` — which also checks *mirror
  presence* — is not documented as offline. Swapping the command would have meant
  asserting something about `audit` I could not verify from its help text. Happy
  to change it if the sweep was the intent.

## Verification

```
uv run pytest -q     → 1723 passed, 25 skipped; coverage 100.00% (gate held)
uv run mypy          → Success: no issues found in 81 source files
uv run ruff check    → All checks passed
pre-commit           → all hooks passed (codespell, detect-secrets, plugin-validate)
```

Before the doc fixes the guard reported exactly the four failures above and
nothing else — 4 failed, 93 passed, 11 skipped.

### DIFF
--- a/defendable-science/tests/test_skill_cli_contract.py
+++ b/defendable-science/tests/test_skill_cli_contract.py
@@ -33,6 +33,7 @@ from __future__ import annotations
 
 import re
 import shlex
+from collections import Counter
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple
 
@@ -75,15 +76,22 @@ _INLINE_START = re.compile(rf"^(?:{_NAMES})\s+(.*)$", re.S)
 #: another command. ``<`` must be followed by whitespace so a ``<placeholder>``
 #: value is not mistaken for an input redirect.
 _SHELL_BREAK = re.compile(r"\s(?:\||&&|;|2?>>?|<<?\s)")
-#: Usage-synopsis grammar — ``[--flag]``, ``(a| b)``, ``x|y``, ``…``. These
-#: describe a command's shape rather than calling it, so arity and flag checks
-#: would both be meaningless.
-_SYNOPSIS = re.compile(r"[\[\]()|…]|\.\.\.")
+#: A blockquote continuation marker. Skills document tooling inside ``>`` quotes,
+#: so a span wrapped across such a line carries a leading ``>`` that would
+#: otherwise read as an output redirect and truncate the invocation.
+_QUOTE_MARKER = re.compile(r"^\s*>\s?", re.M)
+#: Usage-synopsis grammar — ``[--flag]``, ``(a| b)``, ``x|y``. These describe a
+#: command's shape rather than calling it, so arity and flag checks would both be
+#: meaningless. ``...`` counts only as a *standalone* token (``--citekey KEY
+#: ...``); inside a path it is an elision of real directories
+#: (``docs/research/.../strategy.md``) and the call is runnable.
+_SYNOPSIS = re.compile(r"[\[\]()|…]|(?<!\S)\.\.\.")
 #: A placeholder standing in for the command itself: ``defendable-science <group> <cmd>``.
 _TEMPLATE = re.compile(r"^[<{]")
-#: Prose that documents a command's *absence*. Checked against the text just
-#: before an inline span, on the same line.
-_NEGATION = re.compile(r"\b(no|not|never|instead of|rather than)\b[^.]*$", re.I)
+#: Prose that documents a command's *absence*, e.g. "There is no ``X`` command".
+#: Matched against the text between the previous code span and this one, so an
+#: unrelated "not" earlier in the sentence cannot suppress a real invocation.
+_NEGATION = re.compile(r"\b(no|not|never|instead of|rather than)\b[^.`]*$", re.I)
 
 
 class Invocation(NamedTuple):
@@ -136,6 +144,28 @@ def _join_continuations(lines: list[tuple[int, str]]) -> Iterator[tuple[int, str
         yield start, " ".join(p for p in parts if p)
 
 
+def _without_fences(text: str) -> str:
+    """Blank every fenced line, keeping the line count so numbering survives.
+
+    A ```` ``` ```` marker is three backticks, so scanning the raw document for
+    inline spans mis-pairs every span after the first fenced block — and captures
+    a bare fence's whole body as one "span". Both were happening: the guard saw
+    44 of 67 real inline invocations, and whole files yielded none.
+
+    :param text: The document's markdown.
+    :returns: `text` with fenced regions replaced by empty lines.
+    """
+    out: list[str] = []
+    in_fence = False
+    for line in text.splitlines():
+        if _FENCE.match(line):
+            in_fence = not in_fence
+            out.append("")
+            continue
+        out.append("" if in_fence else line)
+    return "\n".join(out)
+
+
 def _invocations(text: str) -> Iterator[Invocation]:
     """Yield every documented invocation in `text`.
 
@@ -147,13 +177,17 @@ def _invocations(text: str) -> Iterator[Invocation]:
         if match := _LINE_START.match(line):
             yield Invocation(number, match.group(1), "fence")
 
-    for match in _INLINE.finditer(text):
-        if not (inner := _INLINE_START.match(match.group(1).strip())):
+    prose = _without_fences(text)
+    for match in _INLINE.finditer(prose):
+        span = _QUOTE_MARKER.sub(" ", match.group(1)).strip()
+        if not (inner := _INLINE_START.match(span)):
             continue
-        before = text[: match.start()].rsplit("\n", maxsplit=1)[-1]
+        # Only back to the previous span on this line: an unrelated negation
+        # earlier in the sentence must not discard a real invocation.
+        before = prose[: match.start()].rsplit("\n", maxsplit=1)[-1]
         if _NEGATION.search(before):
             continue
-        number = text.count("\n", 0, match.start()) + 1
+        number = prose.count("\n", 0, match.start()) + 1
         yield Invocation(number, " ".join(inner.group(1).split()), "inline")
 
 
@@ -295,7 +329,10 @@ def test_documented_invocation_resolves(case: Case) -> None:
         f"call passes {positionals} positional argument(s)"
     )
     arguments = _arguments(command)
-    if arguments and all(p.nargs > 0 for p in arguments):
+    # No `arguments and` guard: `all([])` is True and `sum([])` is 0, so a
+    # command taking *no* positionals is exactly the case that must be checked —
+    # `defendable-science check junk` exits 2, and used to pass this guard.
+    if all(p.nargs > 0 for p in arguments):
         accepted = sum(p.nargs for p in arguments)
         assert positionals <= accepted, (
             f"{where}: `{_name(command)}` takes at most {accepted} positional "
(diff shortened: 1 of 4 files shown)


davorrunje:
## Code review — five defects found and fixed (`c40c59a`)

A review of `790415f` found the guard was checking substantially less than it claimed, and was missing the mirror of the very bug this PR exists to fix. **All five were reproduced before fixing**, and each now has a named regression test.

| # | Site (`790415f`) | Severity | Defect |
|---|---|---|---|
| 1 | `test_skill_cli_contract.py:72` | **High** | Backtick parity shifted across code fences — a third of the inline corpus was invisible |
| 2 | `test_skill_cli_contract.py:298` | Medium | Upper-bound arity check disabled for zero-positional commands |
| 3 | `test_skill_cli_contract.py:81` | Medium | `...` in the synopsis pattern matched inside elided paths |
| 4 | `test_skill_cli_contract.py:86` | Low | Negation rule scanned too much of the preceding line |
| 5 | `test_skill_cli_contract.py:77` | Low | Blockquote `>` continuation read as an output redirect |

---

### 1. Backtick parity shifted across fences — **High**

`_INLINE = re.compile(r"`([^`]+)`", re.S)` ran over the raw document. A ```` ``` ```` marker is three backticks, so the naive pairing consumed them *across* the fence and mis-paired every inline span after the first fenced block in a document. The same root cause fired in reverse: an unlabelled ```` ``` ```` fence was captured as a single "inline span" containing its entire body.

**Measured against a fence-aware extraction: 67 real inline invocations existed, the guard found 44.** `skills/hypothesis-exploration/SKILL.md` and `skills/dataset/SKILL.md` yielded **zero cases each**. Also silently dropped: the whole command-reference table in `docs/guides/literature.md` (including `literature neighbors`), `docs/guides/progress.md`, `skills/research-init/SKILL.md:140`, `skills/digest/SKILL.md:581`. Renaming `literature neighbors` would have left the guard green.

**Fix:** inline spans are now matched against a fence-blanked copy of the document (`_without_fences`), which replaces fenced lines with empty ones so line numbering is unaffected.

### 2. Arity upper bound disabled for zero-argument commands — **Medium**

```python
if arguments and all(p.nargs > 0 for p in arguments):   # before
```

`all([])` is `True` and `sum([])` is `0`, so the leading `arguments and` was the *only* thing suppressing the check — exactly for commands that accept no positionals. Verified: `check junk`, `progress dashboard oops` and `keys list junk` all resolved cleanly through the guard while exiting 2 at runtime with *"Got unexpected extra argument"*.

This is the precise mirror of the `dataset verify` defect this PR fixes: a documented stray argument on a zero-argument command would ship broken.

**Fix:** dropped `arguments and`, with a comment recording why the empty case is the one that matters.

### 3. `...` matched inside elided paths — **Medium**

`_SYNOPSIS` treated any `...` substring as usage grammar, so `--artifact docs/research/.../strategy.md` was classified as a synopsis and skipped. Those were the **only two fenced `defend record` examples in the guides** (`docs/guides/defend.md:27` and `:109`), and `--signed-off-by` occurs nowhere else in the checked corpus — so that option was unvalidated entirely.

**Fix:** `(?<!\S)\.\.\.` — only a standalone `...` token counts as synopsis grammar; inside a path it is an elision of real directories and the call is runnable.

### 4. Negation rule scanned too far — **Low**

`_NEGATION` matched over the entire preceding line fragment, so an unrelated negation earlier in a sentence discarded a real invocation. At `resources/templates/README.md:46` the span `` `defendable-science progress dashboard` `` was dropped because the same line earlier reads *"a stub saying it has **not** been generated"*.

**Fix:** the pattern now stops at the previous code span (`[^.`]*$`), so only text since the last backtick counts.

### 5. Blockquote continuation truncated spans — **Low**

Skills document tooling inside `> **Tooling.**` blockquotes. A span wrapped across such a line carried a leading `> `, which `_SHELL_BREAK` read as an output redirect — so `` `defendable-science backlog\n> park` `` became `backlog > park` and was truncated to `backlog`. Observed at `skills/paper-exploration/SKILL.md:54`; the pattern recurs across skills.

**Fix:** `_QUOTE_MARKER` strips continuation markers before joining a wrapped span.

---

### Corpus recovered

| | before | after |
|---|---|---|
| fenced | 44 | **57** |
| inline | 44 | **62** |
| total | 101 | **119** |

### The vacuity guard was the real failure

`test_the_guard_actually_found_invocations` asserted `>= 40` against 101 actual cases. **That slack is why losing a third of the inline corpus went unnoticed** — a floor loose enough never to need updating is loose enough to hide silent drift.

It is now per-kind (`fence >= 30`, `inline >= 60`, total `>= 95`) and names four skills that must each contribute at least one case, since two of them silently yielded zero before finding #1. That reasoning is written into the test's docstring, not just this comment.

### Verification

```
uv run pytest -q     → 1743 passed, 28 skipped; coverage 100.00% (gate held)
uv run mypy          → Success: no issues found in 81 source files
uv run ruff check    → All checks passed
pre-commit           → all hooks passed
CI                   → green on 3.11 / 3.12 / 3.13 / 3.14
```

The review also confirmed the three documentation fixes are correct, and specifically that `verify <id>` is the right call at `docs/guides/dataset.md:139` rather than a swap to `dataset audit` — `dataset verify`'s `identifier` is `required=True` while `audit`'s is optional, so the two are not interchangeable. That resolves the open question raised in the PR description.

davorrunje:
## Code review, round 2 (high effort) — five more gaps found and fixed (`1b43a8d`)

**No shipped-code bugs this round.** The reviewer cross-checked the guard's verdict for every fenced case against Typer's own vendored Click parser and found **0 mismatches**, so the hand-rolled parsing in `_walk` / `_count_positionals` is faithful to the real parser. It also re-confirmed the four documentation fixes against the live tree.

All five findings were gaps in the guard itself — drift it would have failed to catch. Each reproduced before fixing.

| # | Severity | Gap |
|---|---|---|
| 1 | **Medium** | An invocation downstream of a pipe was never extracted |
| 2 | Low | Required *options* were never checked, only positionals |
| 3 | Low | Negation rule still fired when the negation followed an earlier span |
| 4 | Low | Continuations could leak across a fence boundary |
| 5 | Low | Vacuity floors inconsistently tight |

### 1. Piped invocations invisible — **Medium**

`_LINE_START` anchored the entrypoint to the start of the fenced line, so `docs/guides/keys.md:54` — `echo "$MY_KEY" | defendable-science keys set S2_API_KEY` — yielded **zero cases**, despite being a verbatim shell block under the guide's own convention. Asymmetric with `_SHELL_BREAK`, which already truncated pipelines on the right. Rename `keys set` and that documented command would rot silently.

**Fix:** fenced lines are split on pipeline separators (`_PIPELINE`) before matching, so every segment is checked.

### 2. Required options unchecked — Low

The arity check covered positionals only. Eight commands carry a required option: `defend record --artifact/--target`, `backlog park --provenance`, `backlog add --provenance`, `backlog drop --reason`, `digest extract cells --citekey`, `digest extract record --cells`, `digest depth cells record --cells`. A verbatim example omitting one exits 2 at the point of use — the same failure class as the two `dataset verify` breaks this PR fixes. Nothing in the corpus is broken this way today, so it was latent.

**Fix:** required options are now asserted for `kind == "fence"` calls.

### 3. Negation rule, second ordering — Low

Round 1 tightened `_NEGATION` to stop at the previous backtick, but that only handled one ordering. At `resources/templates/README.md:46` the `not` sits *after* the `init` span's closing backtick, so `progress dashboard` was still silently skipped.

**Fix:** the negation phrase must now sit immediately before the span, with only words and spaces between (`[\w\s]{0,24}$`). Any punctuation in between means the negation is about something else.

### 4. Continuations leaked across fences — Low

`_join_continuations` received every fence's lines flattened together with no separator, so a block whose last line ends in `\` could fold the *next* block's first line into a bogus command reported at the wrong line number.

**Fix:** `_fenced_blocks` yields each block separately and continuations are joined per block.

### 5. Floors inconsistently tight — Low

Round 1 set `inline >= 60` against 62 actual — 2 slack, versus 24 and 27 for the other two. Deleting three inline mentions in an ordinary docs edit would have red-lighted CI with a bare `Counter(...)` for a message.

**Fix:** floors now sit ~15% below the real counts uniformly, and the failure message names both the count when written and what to do:

> `inline invocations dropped to 41, below the 55 floor (63 when written) — if invocations were removed on purpose, lower the floor in this test; if not, the extractor has stopped seeing part of the corpus`

### A round-1 test asserted the wrong thing

`test_an_unrelated_negation_does_not_discard_an_invocation` expected `init` to be **dropped** from *"a stub saying it has **not** been generated by `defendable-science init`"*. But that sentence says `init` **writes** the stub — it does not say `init` is absent — so the span must still be checked. The test had frozen round 1's over-broad negation rule in place as though it were intended. Corrected, with the reasoning in the docstring.

### Corpus

| | initial | round 1 | round 2 |
|---|---|---|---|
| fenced | 44 | 57 | **58** |
| inline | 44 | 62 | **63** |
| total | 101 | 119 | **121** |

### Verification

```
uv run pytest -q     → 1749 passed, 28 skipped; coverage 100.00% (gate held)
uv run mypy          → Success: no issues found in 81 source files
uv run ruff check    → All checks passed
CI                   → green on 3.11 / 3.12 / 3.13 / 3.14
```

Issue #174, opened 2026-08-28
literature: give the enrichment record one authoritative definition


davorrunje:
## Context

`defendable-science literature enrich` and `literature cites` emit what the codebase calls the
"enrichment record" — the stable per-work JSON object a consuming skill reads. It has no single
definition. It is assembled across three functions and described in prose in four more places,
and nothing structurally holds them in agreement:

| Where | What it contributes |
| --- | --- |
| `defendable-science/defendable_science/literature/graph.py:enrich_work` (`:103-116`) | The base keys: `id{openalex,doi,s2,arxiv}`, `title`, `year`, `venue`, `cited_by_count`, `authors`, `abstract` |
| `defendable-science/defendable_science/literature/graph.py:cites` (`:256-259`) | Adds `provenance{source_id,via}` — only on this path |
| `defendable-science/defendable_science/literature/graph.py:enrich` (`:374-400`) | Adds `context_snippet`, `intent`, `is_influential` and, when no S2 key is configured, `degraded` — only under `--context`; also mutates `record["id"]["s2"]` in place |
| `graph.py:94`, `graph.py:240`, `graph.py:363-367` | Three docstrings each describing a different subset of the shape in prose |
| `docs/design/proposals/literature-citation-graph-client.md:63-67,121-122` | The shape restated a fifth time, in a point-in-time design proposal |

The return type at each of these is `dict[str, Any]`, so strict mypy cannot check a key name, a
key's type, or whether a consumer is reading a key that only exists on one of the two commands.
`cli.py:cites` (`:932`) and `cli.py:enrich` (`:964`) both `json.dumps` the dict straight to
stdout, so whatever the assembly produced is the public contract.

Two consequences a consumer actually hits:

1. **The key set differs by command and by flag with no declared discriminator.** `cites` output
   carries `provenance` and never carries `context_snippet`; `enrich --context` carries
   `context_snippet` and never `provenance`; plain `enrich` carries neither. A consumer cannot
   tell "this field is absent because this command does not produce it" from "this field is
   absent because the data was missing" — the same "failed vs. legitimately empty" collapse the
   repo's failure-honesty rule targets.
2. **`degraded` is set on one path only.** `graph.py:382` sets it when no S2 key is configured.
   Nothing declares it as part of the record, so a consumer has no reason to know to look for it.

This was deferred from #169. The design spec for that work,
`docs/superpowers/specs/2026-08-29-pydantic-parsing-boundary-design.md`, considered folding the
[...]

Issue #171, opened 2026-08-28
core: enforce the status-block schema at parse time, not only in `check` (Pydantic for the YAML boundaries)


davorrunje:
## Context

Companion to #169, which scopes Pydantic to the external-input parsing boundary (ADR-0043) and adopts it for the **JSON** boundaries. #169 explicitly defers the YAML ones. This issue covers them.

**Read this first, because it inverts the assumption:** the YAML boundaries are in *substantially better* shape than the JSON ones, and the case for change here is weaker and riskier. Every YAML parse site reviewed on `main` already validates honestly — it distinguishes "unreadable" from "invalid" from "absent", raises a typed domain error naming the file, and carries a remedy:

| Site | What it already does right |
| --- | --- |
| `core/config.py:81-98` | `YAMLError` → `ValueError`; non-mapping → error naming the actual type |
| `literature/registry.py:505-516` | `YAMLError` → `RegistryError` with the path; non-mapping → error naming the type |
| `digest/artifact.py:256-271` | `YAMLError` → `ExtractionError` with path *and* remedy; explicit `isinstance` on `citekey`/`cells` |
| `digest/artifact.py:385-390` | Deliberately centralised — its docstring notes it is "the one place either cell writer's status-block reads touch YAML, so the error is raised in exactly one place rather than duplicated per reader" |
| `progress/collect.py:456-490` | Exemplary: separates unreadable / invalid / missing-key / wrong-type, each with its own finding and remedy. `_milestone` (`:487-500`) renders a non-mapping gate as its own text rather than dropping it — *"a gate nobody can see is a gate nobody meets"* |
| `scaffold/status.py:138-163` | Returns `None` for *absent* and raises `StatusError` for *malformed*, with the docstring stating that `None` "never means valid and empty" |
| `dataset/manifest.py:301-330` | Named `ManifestError` per malformed entry, identifying the offending entry |

[...]

Issue #176, opened 2026-08-29
literature: make WorksPage's results and meta required so a malformed OpenAlex page cannot read as empty


davorrunje:
## Context

`WorksPage` in `defendable-science/defendable_science/literature/graph.py` is the Pydantic
boundary model for one cursor-paginated page of the OpenAlex `/works` endpoint. It was added
by #169 (ADR-0043) to replace hand-rolled `page.get("results", [])` unpacking. Both of its
fields carry defaults:

```python
class WorksPage(ExternalModel):
    results: list[OpenAlexWork] = Field(default_factory=list)
    meta: _PageMeta = Field(default_factory=_PageMeta)
```

`_PageMeta.next_cursor` likewise defaults to `None`.

Those defaults mean a structurally malformed page still validates, and `cites()` treats it as a
legitimate result. Two concrete cases:

| Body | Current behaviour |
| --- | --- |
| `{"results": [<work>]}` — no `meta` at all, mid-pagination | `next_cursor` defaults to `None`, so pagination silently stops and `cites()` returns a **truncated** frontier as if complete |
| `{"error": "Invalid query parameters", "message": "..."}` — an OpenAlex error envelope returned with HTTP 200 | `results` defaults to `[]`, so `cites()` returns `[]` — reported to the caller as "this work has zero citations" |

Both contradict the function's own documented contract. `cites`' docstring states that a
malformed page is a hard error, because *"stopping silently here would return a truncated
frontier as if complete"*, and `parse_work`'s docstring repeats the same reasoning. The
`ExternalModel` base already sets `strict=True`, so the issue is the field defaults, not
coercion.

This is not a regression — the pre-#169 code did `page.get("results", [])` and
`(page.get("meta") or {}).get("next_cursor")`, so the behaviour is identical. But the model was
the place to close it, and #169's final whole-branch review flagged it as the one remaining
place where a malformed payload still passes as a plausible-looking value at the **OpenAlex**
boundary. The S2 boundary is deliberately different — best effort, skip and mark — and is not
in scope here.

## Goal

A structurally malformed OpenAlex `/works` page fails as a hard error naming the source, rather
than being silently read as an empty or final page.

## Where

- `defendable-science/defendable_science/literature/graph.py` — `WorksPage` and `_PageMeta`
  (the model definitions), and `cites`, which consumes them via `parse_obj`
- `defendable-science/tests/test_literature.py` — the `cites` pagination tests; several build
[...]

PR #179, opened 2026-08-29
feat(core): scope Pydantic to the parsing boundary (ADR-0043) and adopt it for JSON (closes #169)

### LINKED ISSUE #169: docs+core: scope Pydantic to the external-input parsing boundary (ADR-0043), then adopt it for JSON parsing
## Context

`CLAUDE.md` currently carries a blanket prohibition:

> **Pydantic is deliberately rejected** (keep the wheel light, no Rust-binary conflicts) — do not reintroduce it.

That rule has **never had an ADR**, yet it is cited as though it did. `decisions/0031-config-driven-cache-dir.md:31-32` reads *"no Pydantic (ADR rejecting it stands)"* — pointing at a document that does not exist. `decisions/0029-api-key-handling.md:84` and `decisions/0038-venue-resolvers-trusted-not-gated.md:76-77` both repeat the prohibition as a decision driver. So a constraint that shaped at least three recorded decisions has no context, no drivers, no rejected alternatives, and nowhere it can be revisited.

The repo owner has since decided that **Pydantic is permitted for parsing YAML and JSON inputs**. This issue records that decision (phase 1) and lands its first adoption across the **JSON** boundaries (phase 2).

Both halves of the original rationale were checked against the repo and neither holds:

1. **"No Rust-binary conflicts"** presupposes the package shares an environment with the consumer's ML stack. It does not, by explicit design — `resources/ensure-tooling.md:56-59` requires an isolated `uv`/`pipx` tool env or a per-user state venv, *"never the consumer repo's project env. This is what lets `defendable-science` depend freely on `typer` / `requests` / `pyyaml` / `pooch` without touching anyone's torch/jax install."* `pydantic-core` cannot collide with a consumer's `torch`/`numpy` because it is never installed beside them.
2. **"Keep the wheel light"** describes a wheel that is already not light. `uv tree --no-dev` on `defendable-science/` at v0.2.2 shows a 13-package runtime tree: `typer` pulls `rich` (→ `markdown-it-py`, `mdurl`, `pygments`), `shellingham`, `annotated-doc`; `pooch` pulls `packaging`, `platformdirs`; `requests` pulls `certifi`, `charset-normalizer`, `idna`, `urllib3`. Adding `pydantic` adds `pydantic-core`, `annotated-types`, `typing-extensions` — a real download-size increase (`pydantic-core` ships a compiled per-platform wheel) but a difference of degree, not of kind.

What the decision is weighed against is the cost of the status quo. The package hand-parses input it did not write at **10 `yaml.safe_load` sites and 10+ `json.loads` sites across 9 modules**, each re-implementing "is the key present / is the value the type we assumed / what now":

| Boundary | Written by | Parsed in |
| --- | --- | --- |
| OpenAlex / Semantic Scholar responses | third-party API | `defendable-science/defendable_science/literature/graph.py`, `literature/acquire.py`, `core/http.py:285` |
| CSL-JSON bibliographies, sidecars | a reference manager, or us | `literature/registry.py:299`, `literature/acquire.py:1939` |
| Croissant dataset manifests | a dataset publisher | `dataset/manifest.py:534`, `cli.py:1549` |
| YAML frontmatter on every artifact | the researcher, by hand | `core/frontmatter.py`, `scaffold/status.py:153`, `digest/artifact.py:259,386`, `progress/collect.py:456`, `literature/registry.py:506` |
| `config.yml`, the key store | the researcher, by hand | `core/config.py:92`, `core/keys.py:172` |

Ad-hoc `data.get(key, default)` is exactly the construct the repo-wide **failure-honesty** rule exists to prevent: a field missing because a response was truncated and a field legitimately absent collapse into the same value at the point of use, so the caller can no longer tell "failed" from "legitimately empty". Separately, an artifact's shape is re-stated in four places that can drift — the writer, the reader, the shipped template under `resources/templates/`, and the rule in `defendable-science/defendable_science/check/checks.py` — with nothing structural holding them in agreement.

## Goal

**Phase 1** — a new ADR records that Pydantic is permitted at the external-input parsing boundary and nowhere else; `CLAUDE.md`'s blanket prohibition is replaced by that scoped rule; every document citing the old prohibition points at the new ADR or is marked as a statement about its own implementation.

**Phase 2** — `pydantic` becomes a pinned runtime dependency, and every **JSON** boundary parses through a validated model that converts a `ValidationError` into the module's existing explicit failure signal. YAML boundaries are explicitly deferred (see Out of scope).

---

# Phase 1 — record the decision

## Where (phase 1)

- **New:** `decisions/0043-pydantic-at-the-parsing-boundary.md` — 0042 is the highest number currently used on `main` (`decisions/0042-depth-mode-matrix-cells-provenance.md`); confirm 0043 is still free before writing, since other branches may have claimed it.
- `decisions/README.md:56` — append the index row after the 0042 row (`| [0043](0043-….md) | <one-line summary> | accepted |`).
- `CLAUDE.md:64` — the "Code style" bullet; replace the "Pydantic is deliberately rejected" sentence.
- `decisions/0031-config-driven-cache-dir.md:31-32` — the dangling *"ADR rejecting it stands"* reference.
- `decisions/0029-api-key-handling.md:84` — "no Pydantic, no dotenv dependency".
- `decisions/0038-venue-resolvers-trusted-not-gated.md:76-77` — "the package deliberately stays light: `requests`, `pooch`, `pyyaml`, no Pydantic".
- `docs/design/proposals/dataset-manifest-tooling.md:46-47,129-130` — "No `jsonschema`, no `pydantic`, no `datasets`".

`decisions/0029-api-key-handling.md:5-7` already establishes the house convention for amending an ADR without rewriting it — a blockquote immediately after the status line (`> **Refined by ADR-0032.** …`). Use that form for 0029/0031/0038 rather than editing their decision text.

## Proposed approach (phase 1)

Write the ADR in the repo's MADR shape (Context · Decision drivers · Considered options · Decision · Consequences · Rejected alternatives), matching the register of `decisions/0041-dashboard-generated-by-the-cli.md`. The context material above is intended to be lifted directly into it.

The five decision points, which are the load-bearing part:

1. **Pydantic is permitted for parsing and validating input the package did not produce.** The boundary is drawn by *provenance*, not by file format: third-party API responses, and file formats read from disk that a human or outside party authored (Croissant manifests, CSL-JSON, YAML frontmatter, `config.yml`, the key store).
2. **Internal value objects remain stdlib `dataclasses`.** Not a general-purpose model layer. Do not convert an existing dataclass that never touches untrusted input, and do not introduce a model for a shape the package both writes and reads within one process.
3. **Where a boundary model exists, it is the authoritative schema for that shape** — the writer, the shipped template and the corresponding `check` rule derive from it or are guarded against it in tests, collapsing the four-way drift to one source. A model that merely duplicates a schema still written longhand elsewhere has not paid for itself.
4. **A validation failure degrades per the failure-honesty rule.** A `ValidationError` is caught at the boundary and converted into the module's existing explicit failure signal — a `Finding`, a non-zero exit, an error object in the JSON envelope — carrying the field path and reason. Never re-raised as a bare traceback, never swallowed into a default. Pydantic makes malformed input *detectable*; this point is what makes it *honest*, and it is the half that matters.
5. **Adoption is incremental and per-boundary**, each slice landing with the tests that cover it. ADR-0028's 100% statement + branch coverage gate is unchanged and applies to the validation and degradation branches too.

Alternatives worth recording as rejected, with why:

- **Keep the blanket rejection** — preserves the lighter wheel (a real cost), but preserves the four-way drift, the hand-rolled parsing at ~20 sites, and three ADRs citing a document that was never written.
- **Pydantic everywhere, replacing `dataclasses`** — buys nothing for objects the package constructs itself, where strict mypy is already the guarantee; costs a rewrite across ~30 modules. An architectural change, not debt repayment.
- **`jsonschema`** — genuinely cheaper on the dependency axis (pure Python, no compiled wheel). Rejected because it validates without *parsing*: the payload is still an untyped `dict`, call sites keep hand-unpacking, mypy learns nothing, and the drift is untouched.
- **`TypedDict` + hand-written validators** — zero new deps, fully mypy-visible; the strongest case against. Rejected as the status quo with a name: those validators *are* the hand-rolled code being re-implemented per boundary.
- **`attrs` + `cattrs`** — comparable ergonomics, pure Python, a close call. Rejected for the narrower ecosystem around JSON-shaped external input, and because Pydantic's structured error objects drop straight into the failure signal point 4 requires, where `cattrs` would need a hand-written translation layer — the very thing being removed.

Consequences worth stating: `requires-python = ">=3.11,<3.15"` (`defendable-science/pyproject.toml`) constrains the acceptable Pydantic version to one publishing wheels across that whole range, which becomes a release-time check; and the obvious risk is scope creep into a general model layer, bounded by point 2 — a reviewer should treat a Pydantic model over package-internal data as a finding.

For `docs/design/proposals/dataset-manifest-tooling.md`, prefer a short amendment note over rewriting the design sketch — it is a point-in-time proposal for already-shipped tooling. Leave `docs/superpowers/plans/*.md` untouched for the same reason: those are dated historical plans (five of them restate "no Pydantic"), and editing them would falsify the record rather than update it.

---

# Phase 2 — add the dependency and adopt it for JSON

## The debt is concentrated, not uniform

**Read this before starting — it determines the order of work and prevents churning code that is already correct.**

Several JSON boundaries already validate honestly: they `isinstance`-check, raise a typed domain error naming the file, and distinguish "malformed" from "absent". `core/keys.py:170-184`, `literature/registry.py:290-307` (`_parse_items`), `cli.py:1805-1823` (`--points`), `cli.py:2383-2395` (`_parse_json_object`) and `cli.py:2774-2800` (`--cells`) are all in this category. For these, Pydantic is a **consolidation** — less boilerplate, one error idiom instead of five — not a bug fix. They are the pattern to generalise *from*, and they go last.

The real defects are concentrated in `literature/graph.py`, `literature/acquire.py` and the Croissant entry point. Each of the following was verified against the code on `main`:

| # | Site | Defect | Symptom |
| --- | --- | --- | --- |
| 1 | `literature/graph.py:256-257` | `for work in page.get("results", [])` passes each element straight to `enrich_work(work)` with no `isinstance` check, though the *page* is checked at `:254` | A non-dict element in `results` raises `AttributeError` on `work.get` — a raw traceback mid-pagination |
| 2 | `literature/graph.py:81-86` (`_abstract`) | `index.items()` assumes `abstract_inverted_index` is a mapping | A string or list value raises `AttributeError` |
| 3 | `literature/graph.py:348-351` (`_aggregate_s2_edges`) | `edge["contexts"][0]` and `edge["intents"][0]` assume a non-empty *list* | If S2 returns a bare string, this silently yields **its first character** as the citation context. A wrong value, no error, straight into the record — the worst failure mode for an integrity tool |
| 4 | `literature/graph.py:110-116` (`enrich_work`) | `publication_year` / `cited_by_count` are passed through with no type check | A string year propagates into the enrichment record and out through emitted CLI JSON |
| 5 | `literature/acquire.py:1939-1943` | `json.loads` then `data["candidate"]`, `data["match"]`, `data["rung"]`, `data["url"]` — bare `KeyError` on a malformed sidecar, plus four `cast()` calls | `cast` is a compile-time assertion with **no runtime check**: a sidecar carrying `rung: 7` passes `cast("str", …)` and flows on as a str-typed int. mypy is actively misled |
| 6 | `cli.py:1549` | `json.loads` result is handed to `manifest_mod.entry_from_croissant(doc)`, which is typed `dict[str, Any]` and immediately calls `json_ld.get("name")` at `dataset/manifest.py:548` | A Croissant file whose top level is a JSON array raises `AttributeError`, which is **not** in the `except (OSError, json.JSONDecodeError, ManifestError)` tuple at `cli.py:1551` — so it escapes as a traceback instead of the documented exit 1 |

Items 3, 5 and 6 are failure-honesty violations in the `CLAUDE.md` sense and should be treated as the justification for the whole phase.

## Where (phase 2)

- `defendable-science/pyproject.toml` — add `pydantic` to `[project].dependencies`; consider enabling the `pydantic.mypy` plugin under `[tool.mypy]`.
- **New:** `defendable-science/defendable_science/core/models.py` (or `core/parsing.py`) — the shared boundary helper and the base model config.
- `defendable-science/defendable_science/literature/graph.py` — defects 1–4.
- `defendable-science/defendable_science/literature/acquire.py:1939-1943` — defect 5.
- `defendable-science/defendable_science/cli.py:1549` + `dataset/manifest.py:534` (`entry_from_croissant`) — defect 6.
- `defendable-science/defendable_science/core/keys.py:170-184`, `literature/registry.py:290-307`, `cli.py:1805-1823`, `cli.py:2383-2395`, `cli.py:2774-2800` — consolidation only, last.
- `defendable-science/defendable_science/core/http.py:155-167,283-289` — see the note on `get_json` below.

## Proposed approach (phase 2)

**Pin.** Add `pydantic>=2.x,<3` to `[project].dependencies`, choosing a 2.x release that publishes wheels for **3.11 through 3.14 inclusive** — `requires-python` is `">=3.11,<3.15"`, so verify the wheel matrix at adoption time rather than taking the latest tag on faith. Run `uv sync && uv lock` and confirm `uv tree --no-dev` shows only the expected three additions.

**One shared seam, written first.** The load-bearing part is decision point 4, not the models. Add a single helper that every boundary funnels through, so the `ValidationError`→domain-error translation exists in exactly one place:

```python
def parse_or_raise[T: BaseModel](
    model: type[T], payload: object, *, source: str, error: Callable[[str], Exception]
) -> T:
    """Validate `payload` against `model`, or raise the caller's domain error.

    :param source: What is being parsed, for the message (a path or URL).
    :param error: The module's own error type — ``RegistryError``, ``ManifestError``,
        ``HttpError``, ``ExtractionError``, …
    :raises Exception: `error`, carrying the field path and reason. Never a
        ``ValidationError`` and never a bare traceback.
    """
```

The message must name the offending field path and the reason, matching the quality already set by `core/keys.py:181-183` and `literature/registry.py:303-306`. Every boundary below calls this; none catches `ValidationError` itself.

**Model config.** Boundary models over third-party payloads should ignore unknown fields (OpenAlex and S2 add fields without notice — `extra="ignore"`, the default). Models over *our own* on-disk formats — the acquire sidecar, the key store — should use `extra="forbid"`, since an unexpected key there means a version mismatch worth surfacing. Do not enable coercion where it would mask a type error: `year` must reject `"2024"` rather than silently accept it, which is `model_config = ConfigDict(strict=True)` on the numeric fields or `StrictInt`.

**Order of work.** Land as separate PRs against this issue:

1. Dependency + `parse_or_raise` + its tests.
2. `graph.py` — models for the OpenAlex work object, the works page (`{results, meta.next_cursor}`), the S2 `externalIds` payload and the S2 citation edge. This kills defects 1–4 together. Note that `enrich_work`'s *output* shape (`:103-116`) is also what the CLI emits, so the same model can serve as the response envelope — worth doing, and it connects to the wider "inconsistent CLI response shapes" problem.
3. `acquire.py` sidecar — a model with required `candidate` / `match` / `rung` / `url` fields deletes all four `cast()` calls and turns the `KeyError` into a named error.
4. Croissant — validate the top level before `entry_from_croissant`, so `cli.py:1551`'s except tuple becomes truthful; keep the existing per-`distribution` checks or fold them into the model.
5. Consolidation of the five already-honest sites onto `parse_or_raise`.

**Leave `HttpClient.get_json` generic.** `core/http.py:187` returns `JsonValue` and is shared by every caller; the model belongs at the call site that knows the expected shape, not in the transport. Likewise the response cache at `core/http.py:155-167` stores raw JSON and should keep doing so — validating on the way out of the cache is what protects against a poisoned entry, and that happens at the call site too.

**Coverage.** ADR-0028's 100% statement + branch gate applies. Each model needs tests for the valid path, the malformed path, *and* the degradation path proving the domain error surfaces with a usable message rather than a `ValidationError` or a traceback. Defects 1–6 should each get a regression test reproducing the exact symptom in the table above.

## Out of scope

**YAML boundaries.** `core/frontmatter.py`, `scaffold/status.py:153`, `digest/artifact.py:259,386`, `progress/collect.py:456`, `literature/registry.py:506`, `core/config.py:92` and `dataset/manifest.py:312` (`datasets.yml`) stay hand-parsed for now. They are author-written rather than machine-generated, and the frontmatter schemas are the ones entangled with `resources/templates/` and `check/checks.py` — collapsing that four-way drift is a larger design task deserving its own issue once the JSON idiom has settled.

## Acceptance criteria

**Phase 1**

- [ ] `decisions/0043-pydantic-at-the-parsing-boundary.md` exists, in MADR form, carrying all five decision points and the rejected alternatives above
- [ ] `decisions/README.md` has the 0043 index row, appended after 0042
- [ ] `CLAUDE.md:64` no longer says Pydantic is rejected; it states the scoped rule (boundary-only, `dataclasses` internally, `ValidationError` → explicit failure signal) and cites ADR-0043
- [ ] `decisions/0031-config-driven-cache-dir.md`'s *"ADR rejecting it stands"* resolves to a real document
- [ ] `decisions/0029-api-key-handling.md` and `decisions/0038-venue-resolvers-trusted-not-gated.md` carry amendment notes in the existing `> **Refined by ADR-00NN.**` form; their decision text is not rewritten
- [ ] `docs/design/proposals/dataset-manifest-tooling.md` carries an amendment note; `docs/superpowers/plans/*.md` are untouched
- [ ] `grep -rn "Pydantic is deliberately rejected\|ADR rejecting it stands" .` returns nothing

**Phase 2**

- [ ] `pydantic` is in `[project].dependencies` in `defendable-science/pyproject.toml`, pinned `>=2,<3` to a release with wheels for 3.11–3.14; `uv lock` updated
- [ ] `uv tree --no-dev` adds only `pydantic`, `pydantic-core`, `annotated-types` (and `typing-extensions`) to the runtime tree
- [ ] A single shared `ValidationError`→domain-error helper exists; `grep -rn "ValidationError" defendable_science/` shows it caught in exactly one module
- [ ] Defect 1: a `results` page containing a non-dict element yields a named `HttpError`, not `AttributeError` — regression test present
- [ ] Defect 2: a non-mapping `abstract_inverted_index` is rejected or ignored explicitly, not an `AttributeError`
- [ ] Defect 3: an S2 edge whose `contexts` is a bare string never yields a single character as the context — regression test asserting the exact old symptom is gone
- [ ] Defect 4: `publication_year: "2024"` is rejected rather than propagated into the enrichment record
- [ ] Defect 5: `literature/acquire.py:1939-1943` has no `cast()` calls left; a sidecar missing `candidate`/`match`/`rung`/`url`, or carrying a wrong type, raises a named error naming the field
- [ ] Defect 6: `defendable-science dataset ingest <file>` on a Croissant file whose top level is a JSON array exits 1 with a message, not a traceback — verify with `printf '[1,2]' > /tmp/x.json && uv run defendable-science dataset ingest /tmp/x.json; echo "exit=$?"`
- [ ] The five already-honest sites route through the shared helper with no loss of message quality (their existing tests still assert the same messages, or the assertions are updated deliberately)
- [ ] `uv run pytest -q` passes with the 100% statement + branch gate intact
- [ ] `uv run mypy` passes strict
- [ ] No Pydantic model is introduced over package-internal data (decision point 2)

## References

- `CLAUDE.md:64` — the prohibition being replaced; `CLAUDE.md:62` — the house rule that material decisions get an ADR
- `decisions/0029-api-key-handling.md:5-7` (amendment-note convention), `:84`
- `decisions/0031-config-driven-cache-dir.md:31-32` (the dangling reference)
- `decisions/0038-venue-resolvers-trusted-not-gated.md:76-77`
- `decisions/0041-dashboard-generated-by-the-cli.md` — register and structure to match
- ADR-0024 (`decisions/0024-tooling-package-and-bootstrap.md`), ADR-0026 (`decisions/0026-independent-versioning-compat-pin.md`) — the isolated-install model the rebuttal rests on
- ADR-0028 (`decisions/0028-100-percent-coverage-gate.md`) — the coverage gate
- `resources/ensure-tooling.md:56-59` — the isolation guarantee
- `docs/design/proposals/literature-citation-graph-client.md` — the design behind `graph.py`
- `docs/design/proposals/dataset-manifest-tooling.md:46-47,129-130`
- `defendable-science/pyproject.toml` — `requires-python`, current runtime dependencies

davorrunje:
## What

`CLAUDE.md` carried a blanket **"Pydantic is deliberately rejected"** rule that had no ADR, yet was cited as a decision driver by three of them — one pointing at a document that was never written. This records the real decision as **ADR-0043** (Pydantic is permitted only where data *enters the process*), then lands its first adoption across the package's JSON boundaries, fixing six verified defects.

Three of the six were failure-honesty violations in the `CLAUDE.md` sense — a malformed third-party payload silently becoming a plausible-looking wrong value, or escaping as a raw traceback, in a tool whose purpose is defensible research.

## Details

**Phase 1 — the decision.** Both halves of the original rationale were re-checked and neither holds. *"No Rust-binary conflicts"* presupposes the package shares an environment with the consumer's ML stack; it does not, by explicit design (ADR-0024/0026, `resources/ensure-tooling.md:56-59`), so `pydantic-core` is never installed beside anyone's torch. *"Keep the wheel light"* describes a 13-package runtime tree. ADR-0029, 0031 and 0038 get amendment blockquotes in the existing `> **Refined by …**` form; their decision text is untouched. `docs/superpowers/plans/*.md` are deliberately left alone as dated historical record.

**The rule, in one line:** a model is warranted for anything the package *reads* that it did not construct in this process. The issue defined the boundary by *provenance*, which does not cover its own examples — the acquire sidecar is written by us and still crosses a process boundary. "Entering the process" needs no carve-out.

**Phase 2 — one seam, then six defects.** `core/models.py` is the only module in the package that imports `ValidationError`; every boundary hands it a `source` and its own error type. The six fixes:

| # | Was | Now |
| --- | --- | --- |
| 1 | a non-dict row in an OpenAlex page → `AttributeError` mid-pagination | named `HttpError`, page hard-fails as one unit |
| 2 | a non-mapping `abstract_inverted_index` → `AttributeError` | rejected at the model |
| 3 | an S2 edge whose `contexts` was a bare string silently yielded **its first character** as the citation context | impossible at the type level; the edge is skipped and the loss marked in `degraded` |
| 4 | `publication_year: "2024"` propagated into the record and out through the CLI | rejected (`strict=True`) |
| 5 | four `cast()` calls on the quarantine sidecar — no runtime check, so `rung: 7` flowed on as a str-typed int | validated `OwnedModel`; named error identifying file and field |
| 6 | a Croissant file whose top level is a JSON array escaped `cli.py`'s `except` tuple as a traceback | exits 1 with a message; the tuple is now truthful, with no CLI change |

**Deliberate asymmetry.** OpenAlex hard-fails a page as one unit — a truncated citation frontier returned as if complete is worse than a loud failure. Semantic Scholar stays best-effort — it is the flakier provider and the enrichment is optional — so a malformed member is skipped and the loss marked in the `degraded` list `enrich` already emits. A late review found the S2 leg had been converted only halfway (a malformed metadata body killed an entire run; a `{"data": null}` citations page raised `TypeError`; a non-dict page returned an *unmarked* empty). All three now degrade and mark.

**Consolidation: two of five sites, not five.** The criterion is "route through the shared helper *with no loss of message quality*", and applied honestly it admits two. `cli.py`'s `_parse_json_object` returns `None` by design and has no domain error to translate to; `--points` and `--cells` still build internal `dataclasses` via `*_from_mapping`, so a container model would duplicate a longhand schema (ADR-0043 point 3) and would cost the `--cells item N` framing the tests assert. Each excluded site now carries a one-line comment so the decision survives.

**Deliberately not done.** Modelling `enrich_work`'s *output* — ADR-0043 point 2 forbids a Pydantic model over data the package only emits; filed as #174. YAML boundaries are out of scope and tracked by #171.

**Dependency.** `pydantic>=2.13,<3`. `uv tree --no-dev` gains five packages, of which only `pydantic-core` is compiled; the other three plus `typing-inspection` are pure Python. The wheel matrix covers all of `requires-python = ">=3.11,<3.15"`.

**Verification.** 1658 tests, 100% statement+branch coverage (ADR-0028 gate intact), `mypy` strict clean across 82 files, `ruff` clean, `validate-plugin.sh` and `pre-commit` green. Each of the six defects has a regression test reproducing its exact original symptom, and every boundary-error assertion pins both the source and the reason.

Closes #169


davorrunje (commented):
First review pass — read every hunk of the source diff (`decisions/`, `docs/`, `uv.lock` skimmed; no issues there). The suite is green from `defendable-science/` (1659 passed, 100.00% stmt+branch), `ruff`/`mypy` clean. The seam itself (`core/models.py`) is well-shaped and the six claimed defects are genuinely fixed and regression-tested.

What I did find is a consistent class of problem in the conversion: **several of the replaced call sites carried an explicit `or {}` / `or []` null guard, and the new models are non-optional with a `default_factory`, which under `strict=True` rejects an explicit JSON `null` rather than treating it as absent.** A `default_factory` only covers a *missing* key, never a `null` one. That silently converts "tolerated null" into "hard error" or "false degradation" at three sites. Each is reproduced below with a probe I ran against this branch.

One further failure-honesty gap survives the conversion (`if meta:` in `_s2_context`), plus three low-severity doc/marker nits.

No blocking objection to the ADR or the architecture — these are all localised.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -34,6 +38,106 @@
...
+class _ExternalIdBundle(ExternalModel):
+    doi: str | None = Field(default=None, alias="DOI")
+    arxiv: str | None = Field(default=None, alias="ArXiv")
+    corpus_id: int | str | None = Field(default=None, alias="CorpusId")
+
+
+class S2ExternalIds(ExternalModel):
+    """A Semantic Scholar paper's ``externalIds`` response."""
+
+    external_ids: _ExternalIdBundle = Field(
```
davorrunje: **Medium — dropped null guard: an explicit `"externalIds": null` now becomes a false `transport_error`.**

The code this replaces in `_s2_crossref` was `ext = (paper.get("externalIds") or {})` — the `or {}` was there specifically because S2 returns `null` for a requested field the paper does not have. `external_ids` here is non-optional with a `default_factory`, and a `default_factory` only covers a **missing** key; a present-but-`null` key still goes through validation, and `strict=True` rejects it.

Verified on this branch:

```python
c = C({f"{S2}/paper/CorpusId:7": {"paperId": "abc", "externalIds": None}})
graph.resolve("CorpusId:7", client=c)
# {'resolved': False,
#  'reason': '...: externalIds: Input should be a valid dictionary ... (got NoneType)',
#  'transport_error': True}
```

So a *well-formed* S2 record for a paper with no DOI/arXiv cross-reference now reports a transport fault (CLI exit 3) instead of the correct miss, `"could not cross-reference S2 id 'CorpusId:7' to a DOI/arXiv"` (exit 1). That is the inverse of the defect #3-family this PR set out to fix: a legitimate negative is now being reported as a failure.

Suggest `external_ids: _ExternalIdBundle | None = Field(default=None, alias="externalIds")` and read it as `ids = parsed.external_ids or _ExternalIdBundle()` at both call sites, preserving the old `or {}` semantics.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -314,15 +463,25 @@ def _s2_context(client: HttpClient, s2_paper_id: str) -> dict[str, Any]:
         raise
     except HttpError:
         meta = {}
-    corpus = (
-        (meta.get("externalIds") or {}).get("CorpusId")
-        if isinstance(meta, dict)
-        else None
-    )
+    corpus = None
+    if meta:
```
davorrunje: **Medium — same dropped null guard, opposite symptom: a false `degraded: ["s2"]`.**

Second consumer of `S2ExternalIds`. The old code was `(meta.get("externalIds") or {}).get("CorpusId")`. With the non-optional field, `{"paperId": "abc", "externalIds": None}` raises, gets caught below, and sets `meta_skipped`:

```python
g._s2_context(C2(), "X")
# {'s2': None, 'context_snippet': 'c', 'intent': 'bg', 'is_influential': True,
#  'meta_skipped': True, 'edges_skipped': 0}
```

`enrich` then emits `record["degraded"] == ["s2"]` for a payload that is not degraded at all — this paper simply has no CorpusId. Over-reporting degradation on healthy data erodes the marker's meaning just as much as under-reporting it.

---

**Medium (separate issue, same line) — `if meta:` skips validation for a malformed *falsy* body, leaving no marker.**

`meta = {}` is assigned on the transport-miss path, so the guard is doing double duty. But `client.get_json` can legitimately return any JSON value on a 200, and every falsy one bypasses the model entirely:

```python
for meta in (None, [], 0, ""):
    g._s2_context(C(meta, {"data": []}), "X")
# -> {'s2': None, ..., 'edges_skipped': 0}   # no meta_skipped, for all four
```

A 200 body of `null` or `[]` from `/paper` is malformed, and it is reported to the consumer as a clean "S2 has no id for this work" with no `degraded` marker — exactly the hole this PR closes one function later with `citations_skipped` (and the same hole the `test_enrich_with_key_non_dict_citations_page_degrades_with_marker` test was added for). Distinguishing "transport miss" from "malformed body" wants a sentinel rather than truthiness — e.g. set `meta = _MISS` in the `except HttpError` branch and test `if meta is not _MISS:`.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -34,6 +38,106 @@
...
+
+class _PageMeta(ExternalModel):
+    next_cursor: str | None = None
+
+
+class WorksPage(ExternalModel):
+    """One cursor-paginated page of the OpenAlex ``/works`` endpoint."""
+
+    results: list[OpenAlexWork] = Field(default_factory=list)
+    meta: _PageMeta = Field(default_factory=_PageMeta)
```
davorrunje: **Low/Medium — dropped null guard: `"meta": null` now aborts the whole citation frontier.**

`cites()` previously ended pagination cleanly via `cursor = (page.get("meta") or {}).get("next_cursor")`. `meta: _PageMeta = Field(default_factory=_PageMeta)` covers a missing key but not a `null` one, and `strict=True` rejects `None`:

```python
# page = {"results": [{"id": ".../W2"}], "meta": None}
graph.cites("W1", client=C3())
# HttpError: https://api.openalex.org/works: meta: Input should be a valid
#            dictionary or instance of _PageMeta (got NoneType)
```

Unlike the `results` case, hard-failing here buys nothing: a `null` meta unambiguously means "no further cursor", and the rows already collected on this page are discarded along with every page before it. `meta: _PageMeta | None = None` with `cursor = page.meta.next_cursor if page.meta else None` restores the old behaviour without weakening the `results` strictness that the truncated-frontier argument actually rests on.

defendable-science/defendable_science/dataset/manifest.py
```diff
@@ -531,30 +534,72 @@ def croissant_for(entry: DatasetEntry) -> dict[str, Any]:
...
+    no gain. ``name`` and ``alternate_name`` are held to ``str | None``
+    instead: the entry **id** is derived from them, and silently coercing a
+    non-string ``name`` into a plausible-looking id is exactly the
+    silently-wrong-value failure ADR-0043 point 4 forbids — a document with a
+    non-string ``name`` must be rejected, not laundered into an id.
+    """
+
+    name: str | None = None
+    alternate_name: str | None = Field(default=None, alias="alternateName")
+    distribution: list[Any] = Field(default_factory=list)
```
davorrunje: **Low/Medium — dropped null guard: `"distribution": null` now fails `dataset ingest`.**

The replaced code was `for i, obj in enumerate(json_ld.get("distribution") or [])` — again an explicit `or []`. Confirmed on this branch:

```python
m.entry_from_croissant({"name": "x", "distribution": None})
# ManifestError: croissant: distribution: Input should be a valid list (got NoneType)
```

`{"name": "x"}` still ingests fine (`test_entry_from_croissant_no_distribution`), so a publisher who writes the key explicitly as `null` is rejected while one who omits it is accepted — a distinction the format does not make. This is a third-party document (`extra="ignore"` is set precisely because publishers do surprising things), and the failure mode is a user-visible `ingest failed:` exit 1 on a Croissant file that worked before this PR.

`distribution: list[Any] | None = None` + `for i, obj in enumerate(doc.distribution or [])` keeps the honest "a *malformed* FileObject is an error" behaviour intact.

Also: `:raises ManifestError:` at line 586 enumerates the rejection cases but does not mention a non-list `distribution`, which is now one of them.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -331,28 +490,47 @@ def _s2_context(client: HttpClient, s2_paper_id: str) -> dict[str, Any]:
...
+        )
+    except HttpError:
+        # A malformed citations page (missing/null/non-list `data`) is the
+        # same "must not hard-fail, must not vanish" situation as a malformed
+        # edge or a malformed metadata body — it must not raise (a `None`
+        # `data` used to blow up as a raw TypeError) and must not report as a
+        # legitimate empty result with no marker.
+        out["citations_skipped"] = True
+        return out
+    out["edges_skipped"] = _aggregate_s2_edges(page.data, out)
```
davorrunje: **Low — `edges_skipped` is set unconditionally, contradicting the docstring contract.**

This assigns the key even when the count is `0`, but the docstring added at line 434-436 states the bookkeeping keys are "never emitted to a consumer directly, and **absent entirely when nothing was lost**". Probe output on a fully healthy call: `{'s2': 'CorpusId:7', ..., 'edges_skipped': 0}`.

Harmless today only because `enrich` happens to test truthiness rather than presence — so the next reader who reaches for `"edges_skipped" in bundle` (the idiom the docstring invites, and the one `meta_skipped`/`citations_skipped` actually use) gets it wrong. Either guard the assignment (`if skipped: out["edges_skipped"] = skipped`) or drop the "absent entirely" sentence.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -396,6 +574,13 @@ def enrich(
                 record["context_snippet"] = bundle["context_snippet"]
                 record["intent"] = bundle["intent"]
                 record["is_influential"] = bundle["is_influential"]
+                degraded: list[str] = []
+                if bundle.get("meta_skipped"):
+                    degraded.append("s2")
+                if bundle.get("edges_skipped") or bundle.get("citations_skipped"):
```
davorrunje: **Low — a single skipped edge marks all three context fields degraded even when they carry real values.**

`/citations` is fetched with `limit=100`. If one of 100 edges is malformed and a surviving edge supplied both a `context_snippet` and an `intent`, this still emits `degraded: ["context", "intent", "is_influential"]` — telling the consumer three populated, correct fields are unreliable.

`citations_skipped` (nothing readable at all) genuinely degrades all three; `edges_skipped` with survivors does not. Worth splitting, since the `degraded` list is a signal a human acts on and this is the noisiest path into it. At minimum, only mark the fields that actually came back `None`.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -171,14 +273,45 @@ def _s2_crossref(client: HttpClient, s2_id: str) -> tuple[str, str] | None:
         raise
     except HttpError:
         return None
-    ext = (paper.get("externalIds") or {}) if isinstance(paper, dict) else {}
-    if ext.get("DOI"):
-        return "doi", str(ext["DOI"])
-    if ext.get("ArXiv"):
-        return "arxiv", str(ext["ArXiv"])
+    ids = parse_obj(
```
davorrunje: **Low — docstring no longer matches: `_s2_crossref` can now raise `HttpError`.**

Before this PR every `HttpError` was caught and converted to `return None`, so `:raises RateLimitError:` was the complete contract. The new `parse_obj(..., error=HttpError)` at line 276 raises out of the function — that is deliberate and `_resolve_s2_xref` handles it, but the docstring still advertises `None` as the only failure outcome. Add `:raises HttpError: If the 200 body is not a well-formed \`externalIds\` response.` so the next caller does not have to read the body to find out.

davorrunje:
## One more, not in the review above: `parse_json` validates in Python mode, so `strict=True` will reject valid JSON

**Latent today, guaranteed to fire on the next boundary.** This is why it isn't in the inline comments — nothing in the current diff triggers it: no boundary model yet declares a `date`, `datetime`, `tuple`, `set`, `UUID` or `Decimal` field. The plan's next three boundaries all do.

### The defect

`core/models.py:113-118`

```python
    try:
        payload = json.loads(text)
    except json.JSONDecodeError as exc:
        raise error(f"{source}: invalid JSON: {exc}") from exc
    return parse_obj(model, payload, source=source, error=error)
```

Decoding first and handing a `dict` to `model_validate` runs validation in **Python mode**. Pydantic deliberately relaxes strictness in *JSON* mode precisely because JSON has no native date, tuple, set, UUID or Decimal — in JSON those are strings and arrays, and that is the only way to express them. In Python mode, `strict=True` correctly insists on a real `datetime.date`, so a perfectly valid payload is rejected.

Reproduced against this branch (`697a96c`) using its own `ExternalModel`, pydantic 2.13.5:

```python
class Work(ExternalModel):
    publication_date: datetime.date

text = '{"publication_date": "2024-01-01"}'

parse_json(Work, text, source="openalex", error=ValueError)
# -> ValueError: openalex: publication_date: Input should be a valid date (got str)

Work.model_validate_json(text)
# -> publication_date=datetime.date(2024, 1, 1)
```

### Why it matters here specifically

The module's opening docstring says a malformed payload must reach the user as "an explicit, actionable failure naming the field and the reason". This produces exactly that — for a **well-formed** payload. A false rejection dressed in a confident error message is the failure-honesty rule running in reverse, and it happens in the one module whose whole job is that rule. `_explain` even renders it convincingly: *"publication_date: Input should be a valid date (got str)"* reads like an upstream bug rather than ours.

The three boundaries named next in the plan each have a date field — OpenAlex `publication_date`, Croissant `dateCreated`, the registry's `fetched` — so the first one to model its date type turns valid input into a hard error.

### Fix

```python
    try:
        return model.model_validate_json(text)
    except ValidationError as exc:
        raise error(f"{source}: {_explain(exc)}") from exc
    # pydantic raises ValidationError for a decode failure too, with
    # type="json_invalid", so the "not JSON" and "wrong shape" cases still
    # share one error idiom — which is what this function exists for.
```

If you prefer to keep the two failure messages distinct (`invalid JSON: …` vs the field path), branch on `err["type"] == "json_invalid"` inside `_explain` rather than pre-decoding.

**`parse_obj` cannot be fixed the same way** — it receives an already-decoded value (from `HttpClient.get_json`, whose cache stores decoded JSON), so it is stuck in Python mode. That asymmetry is worth a line in its docstring: a model with a `date` field validates through `parse_json` but not through `parse_obj`. Otherwise the next author hits it from the other direction.

### Suggested test

```python
def test_parse_json_accepts_json_native_spellings_of_rich_types() -> None:
    """`strict=True` must not reject a date that JSON can only spell as a string."""
    class _M(ExternalModel):
        when: datetime.date

    assert parse_json(_M, '{"when": "2024-01-01"}', source="x", error=ValueError).when == datetime.date(2024, 1, 1)
```

---

Otherwise I agree with the review above, and with its framing: the dominant theme is that a `default_factory` covers a *missing* key while `strict=True` rejects a present-but-`null` one, so the dropped `or {}` / `or []` guards convert tolerated nulls into hard errors. That and this finding are the same underlying shape — the new validation layer is stricter than the data actually is.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 48f7121: `S2ExternalIds.external_ids` is now `_ExternalIdBundle | None`, and `_s2_crossref` reads it as `ids = parse_obj(...).external_ids or _ExternalIdBundle()`, restoring the old `or {}` tolerance. Regression test: `test_resolve_s2_crossref_null_external_ids_is_a_clean_miss`.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 48f7121, same commit as the sibling finding above — `(bundle or _ExternalIdBundle()).corpus_id` at the `_s2_context` call site. Regression test: `test_s2_context_null_external_ids_is_not_meta_skipped`.

Separately fixed the `if meta:` truthiness bug in the same commit, per your second comment on this line — see that reply.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 48f7121: `WorksPage.meta` is now `_PageMeta | None`, and `cites()` reads `page.meta.next_cursor if page.meta else None`. Kept `results` strict, per your note that the truncated-frontier argument rests on that field, not on `meta`. Regression test: `test_cites_null_meta_ends_pagination_without_discarding_the_page`.

Worth flagging: this exact code, plus its sibling for `results`, is also the subject of #176 (filed from the earlier whole-branch review) — that issue asks to make both fields *required* to catch a missing `meta` key as an error too. This fix and that issue are compatible but distinct: this restores tolerance for an explicit `null`; #176 is about a missing key. Left #176 as filed rather than fold it in here.

defendable-science/defendable_science/dataset/manifest.py
davorrunje: Confirmed and fixed in 9e90b3a: `CroissantDocument.distribution` is now `list[Any] | None`, read as `doc.distribution or []`. Extended the `:raises:` docstring to name the non-list case explicitly, per your note. Regression test: `test_entry_from_croissant_null_distribution_ingests_with_no_files`.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 48f7121: `_aggregate_s2_edges`'s skip count is now only assigned to `out["edges_skipped"]` when nonzero. Regression test: `test_s2_context_all_valid_edges_omit_edges_skipped_entirely` asserts the key's absence, not just its falsiness.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 48f7121: `enrich` now checks each of `context_snippet`/`intent`/`is_influential` individually against `None` rather than blanket-marking all three whenever `edges_skipped`/`citations_skipped` is truthy. Two regression tests: `test_enrich_partial_edge_loss_does_not_degrade_recovered_fields` (a surviving edge's real values are not marked degraded) and `test_enrich_total_edge_loss_still_degrades_all_three_fields` (the mirror case, confirming total loss still degrades everything).

defendable-science/defendable_science/literature/graph.py
davorrunje: Fixed in 48f7121 — added `:raises HttpError:` to `_s2_crossref`'s docstring, naming the malformed-body case distinctly from the `None`-return miss.

davorrunje:
Confirmed and fixed in cdc2316 — verified the reproduction first: `model.model_validate_json(text)` accepts `{"publication_date": "2024-01-01"}` where the decode-then-`model_validate` path rejected it.

Went with your first suggested fix (switch to `model_validate_json`, not the `err["type"] == "json_invalid"` branch alone) — but needed the branch in `_explain` too, since JSON mode's decode-failure message has no field path and would otherwise render as `<root>: Invalid JSON: ... (got str)`, breaking the three call sites' `match="invalid JSON"` assertions (case-sensitive substring). `_explain` now special-cases `json_invalid` to render `invalid JSON: <reason>`, preserving that substring.

One thing this surfaced that I hadn't accounted for: JSON mode's wrong-top-level-type message differs from Python mode's (`"Input should be an object"` / `"a valid array"` vs `"a valid dictionary"` / `"a valid list"`). Updated the two existing assertions in `test_keys.py` and `test_lit_registry.py` that pinned the old Python-mode wording — deliberately, not by accident; I confirmed against pre-fix code that both fail without the update and pass with it.

Added your suggested test (`test_parse_json_accepts_json_native_spellings_of_rich_types`) plus the docstring note you flagged on `parse_obj` about the asymmetry.

On the broader framing in your comment: agreed on both counts, and it's the same root cause across the whole review — `default_factory`/pre-decoding cover the *missing* case but not the *present-but-null-or-wire-format* case. Fixed all of it together in 48f7121/9e90b3a; replies on each inline thread with specifics.

One more thing this exercise surfaced that's worth flagging even though it needed no code change: **coverage.py cannot see branch coverage inside a single-line ternary or `or`-fallback expression at all.** I confirmed this with a minimal repro (`x.y if x else None` and `x or {}` both report `num_branches: 0` regardless of whether both paths are exercised). Every null-guard fix in this PR uses exactly that idiom (`bundle or _ExternalIdBundle()`, `page.meta if page.meta else None`, `doc.distribution or []`), which means the 100%-coverage number I reported after the first fix wave was not actually evidence these fixes worked — only my manual REPL checks were. I added explicit regression tests for all of them and confirmed each one fails against the pre-fix code before restoring it, rather than trusting the coverage gate here. Worth keeping in mind for future review of this codebase's many pre-existing `or {}` idioms, since ADR-0028's 100% branch gate has this blind spot repo-wide, not just in this PR.

davorrunje:
Filed #180 for the coverage blind-spot observation — reproduces the gap independent of this PR, surveys ~100 existing single-line ternary/`or`-fallback sites in the package as scope evidence, and lays out four remedy options without picking one. Deferred for later triage, not part of this PR's fix.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -23,17 +23,145 @@
...
+
+    external_ids: _ExternalIdBundle | None = Field(default=None, alias="externalIds")
+
+
+class S2CitationEdge(ExternalModel):
+    """One incoming citation edge from S2's ``/citations`` endpoint."""
+
+    contexts: list[str] = Field(default_factory=list)
+    intents: list[str] = Field(default_factory=list)
+    is_influential: bool = Field(default=False, alias="isInfluential")
```
davorrunje: **Low — an explicit `"isInfluential": null` now discards the whole edge, losing its context and intent too.**

`is_influential: bool = Field(default=False, alias="isInfluential")` under `strict=True` rejects `None`, so `parse_each` drops the entire edge — not just the flag. Verified on this branch:

```python
_aggregate_s2_edges([{"contexts": ["c"], "intents": ["bg"], "isInfluential": None}], out)
# -> skipped=1, out all-None
```

Before this PR, `edge.get("isInfluential")` returned `None`, was falsy, and the edge's `contexts`/`intents` were still harvested.

This is the same null-vs-missing asymmetry already fixed in this round for `WorksPage.meta`, `S2ExternalIds.external_ids` and `CroissantDocument.distribution` — `bool | None = None` restores the prior best-effort behaviour with the truthiness check at line 570 unchanged.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -396,6 +615,24 @@ def enrich(
                 record["context_snippet"] = bundle["context_snippet"]
                 record["intent"] = bundle["intent"]
                 record["is_influential"] = bundle["is_influential"]
+                degraded: list[str] = []
```
davorrunje: **Low (pre-existing, but newly conspicuous) — "S2 was never asked" is reported as "S2 returned nothing".**

When `_s2_paper_id(record)` returns `None` at line 603 — a work with neither a DOI nor an arXiv id — the fallback bundle at 604-612 sets all three context fields to `None`. Those are non-`None`-free but there is no `meta_skipped`/`edges_skipped`/`citations_skipped` key, so `degraded` stays empty and no marker is written.

`enrich`'s own docstring defines "null without the marker" as "S2 was queried and returned nothing". Here S2 was never queried at all, so `literature enrich --context W123` on an OpenAlex-only work reports a confident "S2 had nothing to say" about a work S2 was never asked about — the failed-vs-legitimately-empty distinction the repo treats as load-bearing.

Not introduced by this diff, but the new marker discipline immediately below makes the gap visible. A `not_addressable` marker (or adding the fields to `degraded` in the `s2_id is None` branch) would close it.

davorrunje (commented):
## Review round 2 — `9e90b3a5`

Re-reviewed at high effort after the three fix commits. I re-verified the load-bearing Pydantic behaviour empirically against `pydantic==2.13.5` (strict `RootModel`, lax `list[Any]` rejecting a dict/str/null body, `json_invalid` carrying `ctx["error"]`, alias + `extra="ignore"`, `externalIds: null`) — all hold. The quarantine sidecar's `extra="forbid"` matches `_write_quarantine`'s exact four keys, `RetrievalError` from the new sidecar validation is already caught at `cli.py:1360`, and the lock file carries cp311–cp314 `pydantic-core` wheels matching `requires-python = ">=3.11,<3.15"`.

**Six of the seven round-1 threads are fixed and verified in source with the named regression tests — resolving those.** The seventh (partial edge loss) is fixed as written but the new per-field check opened an adjacent gap; I've replied in that thread rather than opening a new one.

Two further findings inline, both low, both in the same S2 degradation area.

### Not raised as findings
- `OpenAlexWork.ids` is `Field(default_factory=_WorkIds)` without `| None` — the one remaining member of the explicit-null family (`WorksPage.meta`, `S2ExternalIds.external_ids`, `CroissantDocument.distribution` all got the treatment) that would hard-fail an entire page on `"ids": null`. OpenAlex does not appear to send it, so this is a note, not a request.
- The PR body still says the runtime tree "gains five packages"; ADR-0043 was corrected to four in `89a6557`. Body text only, no code impact.

defendable-science/defendable_science/literature/graph.py
davorrunje: Verified fixed for the case reported — but the per-field check opened the mirror-image gap, so leaving this thread open.

**Medium — a partially-skipped page now emits _no_ marker at all, and `is_influential: false` can be plain wrong.**

`context_snippet` and `intent` are *representative samples* — one surviving edge legitimately supplies them, which is exactly why blanket-degrading them was wrong. `is_influential` is not: line 570-573 computes it as `any(edge.isInfluential)` over **all** edges, so a dropped edge can flip it from `true` to `false`.

Concrete scenario — `/citations` returns two edges, the first malformed and influential, the second well-formed and not:

```python
[{"contexts": "an influential mention", "intents": ["bg"], "isInfluential": True},   # skipped
 {"contexts": ["c"], "intents": ["bg"], "isInfluential": False}]                     # kept
```

`_aggregate_s2_edges` skips edge 1, harvests context/intent from edge 2, and line 572 sets `is_influential = False`. `_s2_context` sets `edges_skipped: 1`. In `enrich` all three fields are non-`None`, so the loop at 627 appends nothing, `if degraded:` at 634 is false, and the key is never written. The consumer gets `is_influential: false` with no signal that anything was lost — the truth is `true`.

That contradicts `_aggregate_s2_edges`'s own docstring ("the count is returned so the caller can mark the loss instead of hiding it") and is the same silently-wrong-value class as the round-1 defect this PR set out to fix. `test_enrich_partial_edge_loss_does_not_degrade_recovered_fields` doesn't catch it because its skipped edge isn't influential; `test_enrich_marks_degraded_when_an_s2_edge_is_skipped` covers only the all-edges-skipped case, where `is_influential` is `None` and the marker does fire.

Suggested fix: treat `is_influential` as degraded whenever `edges_skipped`/`citations_skipped` is set (it is an aggregate, not a sample), keeping the per-field check for the two sampled fields — or emit `edges_skipped` itself in the record.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 954ca26: `S2CitationEdge.is_influential` is now `bool | None = None`, matching the old `edge.get("isInfluential")` truthiness. Regression test: `test_s2_edge_with_null_is_influential_keeps_its_context_and_intent` — asserts the edge survives (skipped=0) and its `contexts`/`intent` are harvested.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 954ca26. There was actually a pre-existing test (`test_enrich_with_key_no_addressable_id`) that explicitly pinned this as correct behavior — updated its assertion and docstring to the fixed behavior rather than adding a duplicate test. The fallback bundle now carries `not_addressable: True`, which `_s2_degraded_fields` (see reply below) treats the same as a total loss.

defendable-science/defendable_science/literature/graph.py
davorrunje: Confirmed and fixed in 954ca26, same commit. Went with your suggested split: `context_snippet`/`intent` keep the per-field null check (a surviving edge can still supply a real sample), but `is_influential` is now always marked degraded whenever `edges_skipped`/`citations_skipped`/`not_addressable` is set, regardless of the value it landed on — it's an aggregate, not a sample, so there's no value of it that's trustworthy after any loss.

Extracted the whole decision into `_s2_degraded_fields` — `enrich` tripped ruff's mccabe complexity threshold once the third marker was added, and the decision reads better as its own testable unit anyway.

Reproduced your exact scenario as a regression test (`test_enrich_marks_is_influential_degraded_even_when_samples_survive`) and confirmed it fails without the fix: pre-fix, `rec["is_influential"] is False` with no `degraded" key at all — the silently-wrong-value case you called out. Also updated `test_enrich_partial_edge_loss_does_not_degrade_recovered_samples` (renamed from ...`_fields`), which had pinned the old blanket-exempt behavior for `is_influential" too, once its skipped edge happened not to be the influential one.

defendable-science/defendable_science/literature/graph.py
```diff
@@ -23,17 +23,154 @@
...
+    ``is_influential`` is ``| None``, not merely defaulted: S2 sends an
+    explicit ``null`` when it has no opinion on an edge, and a
+    ``default_factory``/``default`` alone only covers the *missing* key —
+    ``strict=True`` would reject a present ``null``, dropping the *whole*
+    edge (its ``contexts``/``intents`` too) rather than just the flag.
+    ``None`` is falsy at the one call site that reads it, matching the old
+    ``edge.get("isInfluential")`` behaviour.
+    """
+
+    contexts: list[str] = Field(default_factory=list)
```
davorrunje: **Low — `contexts` and `intents` are the last two fields on this model that still drop the whole edge on an explicit null.**

`is_influential` just got the `| None` treatment for exactly this reason; its two siblings keep `Field(default_factory=list)`, which covers a *missing* key but not a present `null`. Probed against `pydantic 2.13.5` with this model's config (`strict=True, extra="ignore"`):

```
FAIL {'contexts': None, 'intents': ['background'], 'isInfluential': True}  -> list_type ('contexts',)
OK   {'intents': ['background'], 'isInfluential': True}                   -> contexts=[]
FAIL {'contexts': ['c'], 'intents': None, 'isInfluential': True}          -> list_type ('intents',)
OK   {'contexts': ['c'], 'intents': ['bg'], 'isInfluential': None}        -> the new fix
```

So a null `contexts` loses the edge's real `intents` and its influence flag, and vice versa.

Two caveats that keep this low, not medium:

1. **The loss is marked, not silent** — `parse_each` counts the skip, so `edges_skipped` fires and the fields degrade. This is avoidable data loss, not a wrong value presented as right.
2. **I have not observed S2 emitting `"contexts": null`.** The observed null-for-an-absent-requested-field case was `externalIds` on `/paper`. This is consistency insurance across the family (`meta`, `external_ids`, `distribution`, `isInfluential` all took it), argued from the same rationale this commit's own docstring gives — not a reproduced production failure.

Worth noting the cost went *up* slightly with this commit: a dropped edge now degrades `is_influential` for the whole work, where before it only degraded fields that came back `None`.

Fix if you take it: `list[str] | None = None` on both, read as `edge.contexts or []` / `edge.intents or []` at `_aggregate_s2_edges`. Reasonable to decline as speculative — S2 not sending it makes this dead defensive code, and the loss is already visible in `degraded`.

davorrunje (commented):
## Review round 3 — `954ca26`

All three round-2 findings verified fixed in source, each with a test that reproduces the exact reported scenario (the two-edge influential/non-influential case is reproduced verbatim). CI is green on 3.11–3.14 with the 100% branch-coverage gate, so the new `_s2_degraded_fields` branches are covered.

- `S2CitationEdge.is_influential` → `bool | None = None`; the one read site (`graph.py:579`) is a truthiness check, so `None` behaves exactly like the old `edge.get("isInfluential")`. ✔
- `not_addressable: True` on the `s2_id is None` fallback bundle, marked degraded in `_s2_degraded_fields`; the pre-existing test that pinned the bug as correct behaviour was flipped. ✔
- `is_influential` now always degraded when anything was skipped, with `context_snippet`/`intent` keeping the per-field check. The docstring on `_s2_degraded_fields` states the sample-vs-aggregate distinction explicitly, which is the right place for it. ✔

Extracting `_s2_degraded_fields` was the right call — the three loss causes now read as three named cases instead of nested conditionals.

One new low finding inline, the last member of the null family on the same model.

### Notes (not findings, no action expected)
- `OpenAlexWork.ids` (`graph.py:76`) is still `Field(default_factory=_WorkIds)` without `| None` — carried over from round 2. Unlike the S2 edge case this one hard-fails the whole work, and inside `WorksPage.results` the whole page. I have no evidence OpenAlex sends `"ids": null`, so it stays a note.
- `docs/design/proposals/literature-citation-graph-client.md:69` still describes the marker as `degraded: ["context","intent"]` for the OpenAlex-only case. The code emits three fields and now has three distinct triggers (`no key`, `not_addressable`, `edges/citations skipped`). The file is outside this PR's diff, so this is drift to pick up separately rather than here.

defendable-science/defendable_science/literature/graph.py
davorrunje: Taken rather than declined, though your framing of it as optional was fair.

Verified the probe reproduces exactly — a null `contexts` loses the edge's real `intents` and flag, and vice versa. Fixed in cc2cada: both are now `list[str] | None = None`. No call-site change was needed after all; `_aggregate_s2_edges` already guards with truthiness (`if ... and edge.contexts`), so `None` short-circuits exactly as `edge.get("contexts")` did.

Two reasons I took it over declining, the second of which is yours:

1. There is no principled reason for these two fields to differ from the four siblings that already got the treatment. `WorksPage.results` vs `meta` *does* have one — results stays strict because the truncated-frontier detection rests on it. Here there is none, and an unexplained 4-of-6 split is what the next author trips on.
2. Your note that the cost went up with the previous commit. A dropped edge now degrades `is_influential` for the whole work, so the expected cost of dropping one is strictly higher than when you first weighed this.

On your caveat about dead defensive code: agreed it is unobserved rather than reproduced, and the commit message says so rather than implying a production failure.

One thing I checked explicitly, since tolerating null could plausibly have loosened the wrong thing: a bare-string `contexts` is **still** rejected (skipped=1) — the defect-3 guarantee this PR opened with. Pinned by its own test next to the two null cases, so a future widening of these fields cannot quietly take it out.

davorrunje: Verified fixed by `cc2cada` — resolving.

`contexts` and `intents` are now `list[str] | None = None` (`graph.py:142-143`) and `_aggregate_s2_edges` reads them by truthiness (`:579-582`), so a null keeps the rest of its edge. Re-ran the four probe cases from the comment above against `cc2cada`, and all four now survive with their remaining fields intact:

```
KEPT  skipped=0  {contexts: None,  intents: [background], isInfluential: True}  -> intent=background infl=True
KEPT  skipped=0  {                 intents: [background], isInfluential: True}  -> intent=background infl=True
KEPT  skipped=0  {contexts: [c],   intents: None,         isInfluential: True}  -> snippet=c        infl=True
KEPT  skipped=0  {contexts: [c],   intents: [bg],         isInfluential: None}  -> snippet=c intent=bg infl=False
```

The first and third were the failures; both now keep the edge. Note this also removes the cost I flagged at the end — a null in one field no longer degrades `is_influential` for the whole work.

PR #183, opened 2026-08-29
fix(core): guard identifier-taking paths against traversal (closes #182)

### LINKED ISSUE #182: core: guard Layout's identifier-taking paths against traversal (citekey/paper_id/slug)
## Context

The codebase audit's critical gap 5 (`.codebase_audit/executive-summary.md`, BE-1 in
`.codebase_audit/be-tech-debt.md`) found that externally-derived identifiers reach the
filesystem unchecked at the `Layout` boundary. Reproduced against `main`:

```python
from pathlib import Path
from defendable_science.scaffold.layout import Layout

r = Path("/tmp/dsprobe")
lay = Layout(repo_root=r, research_root=r / "research", literature_dir=r / "lit",
             datasets_manifest=r / "datasets.yml", thesis_dir=r / "thesis")

p = lay.digest("../../../../../../tmp/dsaudit/PWNED")
p.resolve()  # -> /tmp/dsaudit/PWNED.md — escapes repo_root entirely
```

`citekey` at this call site comes straight from a `--citekey` CLI option
(`cli.py:2751`, `:2951`, `:3103`, `:3252`, `:3302`, `:3590`) with no validation before
`Layout.digest` joins it onto a path. `paper_id` has the same shape at
`exploration/backlog.py:598,604`, `check/checks.py:614,635`, `cli.py:2010,2260,2625`, and
`progress/collect.py:144` — sourced from CLI options and from `row["id"]` in backlog/registry
data, neither of which is validated as a filesystem-safe token before use.

**The rule already exists in this codebase, just not applied here** — this is an
inconsistency, not a missing capability:

- `literature/acquire.py:1035`'s `_safe_name(citekey)` reduces a citekey to
  `[A-Za-z0-9._-]` before it touches the cache filesystem (`:1056`, `:1472`, `:2014`), with
  exactly this rationale: *"a realistic one is already safe and passes through unchanged. One
  containing a path separator would otherwise write outside the cache, which is not a risk
  worth carrying for a field nobody validates."*
- `scaffold/layout.py:216`'s `_relative` and `cli.py:359`'s inline copy of the same check
  (explicitly cross-referenced in a comment: *"Same containment rule as `_relative` in
  scaffold/layout.py"*) both resolve-and-check-containment for **paths read from
  `config.yml`** — a different threat shape (a configured relative path, not a bare
  identifier), but proof the containment idiom is already established in this file.

Neither guard is applied to the identifier-taking `Layout` methods.

This is the same pattern as the BE-4 finding fixed in #169/#179: a validation rule was
written once, applied to one reader, and not generalised to the sibling reader that needed
it just as much.

## Goal

No `Layout` method that takes an identifier (`citekey`, `paper_id`, `slug`) can be made to
resolve outside `repo_root`, regardless of what the caller passes — reject at the point the
identifier is used, with a named error rather than a silent write outside the work tree.

## Where

- `defendable-science/defendable_science/scaffold/layout.py` — the vulnerable methods:
  `digest` (`:144`), `paper_dir` (`:179`), `backlog` (`:183`, via `paper_dir`),
  `hypotheses_dir` (`:187`, via `paper_dir`), `paper_docs_dir` (`:191`, via `paper_dir`),
  `positioning` (`:195`, via `paper_docs_dir`), `hypothesis_dir` (`:199`, both `paper_id` and
  `slug`). `paper_dir` is the common root for five of the seven, so fixing it closes most of
  the surface in one place.
- `defendable-science/defendable_science/defend/record.py:320` — `append_log_entry`, which
  takes an externally-influenced `stem` for the same reason (the audit's finding also names
  this site)
- `defendable-science/defendable_science/literature/acquire.py:1035` — `_safe_name`, the
  existing pattern to reuse or generalise rather than re-invent
- `defendable-science/defendable_science/scaffold/layout.py:216` — `_relative`, the existing
  resolve-and-check-containment idiom (different threat shape, same file, useful precedent
  for the error-message register)

## Proposed approach

Not prescribed in detail — this needs a short design before implementation (bounded task:
the flow exists, the pattern to reuse exists, the fix is localised). The shape most consistent
with the codebase's existing choices:

- A single validation function, likely living in `scaffold/layout.py` next to `_relative` (or
  promoted to `core/` if `defend/record.py` needs it too — `append_log_entry`'s `stem` is the
  same class of untrusted identifier), applied to `citekey`/`paper_id`/`slug` before they are
  joined onto a path.
- Two design choices to settle explicitly, since the two existing precedents disagree:
  **allow-list by charset** (`_safe_name`'s approach — reduce to `[A-Za-z0-9._-]`, silently
  coercing) versus **reject with a named error** (`_relative`'s approach — raise on an escape
  attempt). `_safe_name`'s docstring argues coercion is fine because "a realistic [citekey] is
  already safe and passes through unchanged" — but `Layout` is a lower-level, more widely used
  boundary than the acquisition cache, and a silent coercion here could quietly redirect a
  paper's artifacts to the wrong directory rather than failing loudly. A named-error rejection
  seems more consistent with the repo's failure-honesty rule, but this is exactly the kind of
  call worth a two-sentence design and an explicit yes before implementing.
- `hypothesis_dir`'s `slug` parameter currently has no live caller in the package (only
  `_slug()` in `exploration/backlog.py`, which already sanitises to `[a-z0-9-]`, produces
  slugs today) — worth guarding for defense-in-depth and consistency, but it is not part of
  the reproduced exploit.

## Acceptance criteria

- [ ] `Layout.digest("../../../../../../tmp/dsaudit/PWNED")` (and the equivalent for
      `paper_dir`, `positioning`, `backlog`, `hypothesis_dir`) does not resolve outside
      `repo_root` — verify by reproducing the exact probe above and asserting on the outcome
      (error raised, or coerced-and-contained — per the design decision)
- [ ] `defend/record.py`'s `append_log_entry` is covered by the same guard for its `stem`
      parameter
- [ ] The fix does not duplicate `_safe_name` or `_relative` without a stated reason; if a
      third implementation of "make this filesystem-safe" is introduced, the issue/PR says why
      the existing two were not reused
- [ ] A regression test reproduces the exact traversal string from this issue and asserts the
      new behavior

## References

- `.codebase_audit/executive-summary.md` — critical gap 5
- `.codebase_audit/be-tech-debt.md` BE-1 — the original finding and reproduction
- #169/#179 — the BE-4 fix, the same "rule exists in one reader, not the sibling" shape
- `defendable-science/defendable_science/literature/acquire.py:1035` — `_safe_name`
- `defendable-science/defendable_science/scaffold/layout.py:216` — `_relative`
- `defendable-science/defendable_science/cli.py:359` — the second, cross-referenced copy of
  the containment check

davorrunje:
## What

Closes gap 5 of the codebase audit: an externally-derived identifier (`--citekey`, a
backlog/registry `paper_id`) reached the filesystem unchecked, letting a traversal
attempt write outside the repository entirely.

## Details

**Reproduced against `main` before the fix:**

```python
Layout.digest("../../../../../../tmp/dsaudit/PWNED").resolve()
# -> /tmp/dsaudit/PWNED.md — escapes repo_root entirely
```

**Not a missing capability — an inconsistency.** `literature/acquire.py`'s `_safe_name`
already sanitizes a citekey before it touches the acquisition cache. The rule was applied to
one reader and not the sibling reader (`Layout`) that needed it just as much — the same shape
as the JSON-boundary fix in #169/#179.

**New shared guard, `core/paths.py`.** `require_path_segment(value, *, what, error)` rejects
anything that isn't a single filesystem path segment (empty, `.`/`..`, or containing a
separator), raising the caller's own error type — mirrors the `core/models.py` pattern this
repo already established for the JSON parsing boundary (`error=SomeErrorType`), so no bare
exception leaks across a module.

**Reject, not coerce — a deliberate choice, not a default.** `_safe_name` coerces (replaces
bad characters, always succeeds); `_relative` (the existing `config.yml` path check) rejects.
Went with reject: `Layout` is a lower-level, more widely used boundary than the acquisition
cache, and silently coercing a bad identifier could quietly redirect a paper's artifacts into
the wrong directory with no error — worse than a loud failure, for a tool whose whole point is
a trustworthy record.

**Applied at:**
- `Layout.digest` (citekey)
- `Layout.paper_dir` (paper_id) — the shared root of `backlog`/`hypotheses_dir`/
  `paper_docs_dir`/`positioning`, so this one guard covers all four
- `Layout.hypothesis_dir` (slug, in addition to paper_id via `paper_dir`)
- `defend/record.py`'s `append_log_entry` (stem) — the other site the audit named

**Left alone, deliberately:** `acquire.py`'s `_safe_name` keeps its existing coercion for the
acquisition cache — a separate, already-working boundary, not part of this fix.

Every regression test reproduces the exact traversal string from the issue and confirms it
fails against pre-fix code before the guard was added.

Closes #182

### DIFF
--- a/defendable-science/defendable_science/core/paths.py
+++ b/defendable-science/defendable_science/core/paths.py
@@ -1,0 +1,41 @@
+"""Guard for externally-derived identifiers joined onto a filesystem path.
+
+A citekey, paper id, or accountability-log stem is a single path *segment* —
+never a sub-path — but each arrives from outside the process (a CLI option, a
+CSL-JSON field, backlog data) with nothing checked. Joining one unvalidated
+onto a path lets ``..`` or an absolute-looking value escape the directory it
+was meant to stay inside (defendable-science#182). This is deliberately
+stricter than :func:`~defendable_science.scaffold.layout._relative`, which
+resolves and checks containment for a *configured path* that may legitimately
+have subdirectories — an identifier never should.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+_UNSAFE = frozenset({"", ".", ".."})
+
+
+def require_path_segment(
+    value: str, *, what: str, error: Callable[[str], Exception]
+) -> str:
+    """Reject `value` unless it is safe as a single filesystem path segment.
+
+    :param value: The externally-derived identifier (citekey, paper id, log stem).
+    :param what: What is being validated, for the error message (e.g. ``"citekey"``).
+    :param error: The calling module's own error type (e.g. ``LayoutError``,
+        ``RecordError``) — never a bare ``ValueError`` leaking past the caller's
+        module boundary.
+    :returns: `value` unchanged.
+    :raises Exception: `error`, naming `what` and the offending value, if `value`
+        is empty, ``.``/``..``, or contains a path separator — anything that
+        would make it address more than one path segment and let it escape the
+        directory it is joined into.
+    """
+    if value in _UNSAFE or "/" in value or "\\" in value:
+        raise error(f"{what} is not a valid path segment: {value!r}")
+    return value
--- a/defendable-science/defendable_science/core/paths.py
+++ b/defendable-science/defendable_science/core/paths.py
@@ -32,10 +32,11 @@ def require_path_segment(
         module boundary.
     :returns: `value` unchanged.
     :raises Exception: `error`, naming `what` and the offending value, if `value`
-        is empty, ``.``/``..``, or contains a path separator — anything that
-        would make it address more than one path segment and let it escape the
-        directory it is joined into.
+        is empty, ``.``/``..``, contains a path separator, or contains a control
+        character (e.g. an embedded NUL) - anything that would either address
+        more than one path segment or reach the OS call as a raw, uncaught
+        error instead of this function's own signal.
     """
-    if value in _UNSAFE or "/" in value or "\\" in value:
+    if value in _UNSAFE or "/" in value or "\\" in value or any(c < " " for c in value):
         raise error(f"{what} is not a valid path segment: {value!r}")
     return value
--- a/defendable-science/defendable_science/core/paths.py
+++ b/defendable-science/defendable_science/core/paths.py
@@ -8,10 +8,18 @@ was meant to stay inside (defendable-science#182). This is deliberately
 stricter than :func:`~defendable_science.scaffold.layout._relative`, which
 resolves and checks containment for a *configured path* that may legitimately
 have subdirectories — an identifier never should.
+
+The backslash check makes Windows nominally in scope, so a Windows *drive*
+(``"C:PWNED"``, drive-relative — no separator at all, so escapes the other
+checks) is rejected too, via :mod:`ntpath` rather than :mod:`os.path`: the
+latter is a no-op for drive detection on the POSIX hosts this test suite
+actually runs on, which would make the check silently do nothing everywhere
+it is exercised.
 """
 
 from __future__ import annotations
 
+import ntpath
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -32,11 +40,18 @@ def require_path_segment(
         module boundary.
     :returns: `value` unchanged.
     :raises Exception: `error`, naming `what` and the offending value, if `value`
-        is empty, ``.``/``..``, contains a path separator, or contains a control
-        character (e.g. an embedded NUL) - anything that would either address
-        more than one path segment or reach the OS call as a raw, uncaught
-        error instead of this function's own signal.
+        is empty, ``.``/``..``, contains a path separator, contains a control
+        character (e.g. an embedded NUL), or is a Windows drive-relative value
+        (``"C:PWNED"``) — anything that would either address more than one
+        path segment or reach the OS call as a raw, uncaught error instead of
+        this function's own signal.
     """
-    if value in _UNSAFE or "/" in value or "\\" in value or any(c < " " for c in value):
+    if (
+        value in _UNSAFE
+        or "/" in value
+        or "\\" in value
+        or any(c < " " for c in value)
+        or ntpath.splitdrive(value)[0]
+    ):
         raise error(f"{what} is not a valid path segment: {value!r}")
     return value
--- a/defendable-science/defendable_science/defend/record.py
+++ b/defendable-science/defendable_science/defend/record.py
@@ -339,10 +339,13 @@ def append_log_entry(log_dir: Path, date: str, stem: str, body: str) -> Path:
         naming scheme changes.
     :param body: The rendered YAML to write.
     :returns: The path written.
-    :raises RecordError: If `stem` is not a single path segment — see
-        :func:`~defendable_science.core.paths.require_path_segment`
-        (defendable-science#182).
+    :raises RecordError: If `date` or `stem` is not a single path segment —
+        see :func:`~defendable_science.core.paths.require_path_segment`
+        (defendable-science#182). No current caller passes an untrusted
+        `date` (both are machine-generated), but this is public API and the
+        guard covers the whole filename it builds, not half of it.
     """
+    date = require_path_segment(date, what="date", error=RecordError)
     stem = require_path_segment(stem, what="stem", error=RecordError)
     log_dir.mkdir(parents=True, exist_ok=True)
     target = log_dir / f"{date}-{stem}.yml"
--- a/defendable-science/defendable_science/defend/record.py
+++ b/defendable-science/defendable_science/defend/record.py
@@ -26,6 +26,7 @@ from defendable_science.core.frontmatter import (
     set_field,
     split_frontmatter,
 )
+from defendable_science.core.paths import require_path_segment
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -338,7 +339,11 @@ def append_log_entry(log_dir: Path, date: str, stem: str, body: str) -> Path:
         naming scheme changes.
     :param body: The rendered YAML to write.
     :returns: The path written.
+    :raises RecordError: If `stem` is not a single path segment — see
+        :func:`~defendable_science.core.paths.require_path_segment`
+        (defendable-science#182).
     """
+    stem = require_path_segment(stem, what="stem", error=RecordError)
     log_dir.mkdir(parents=True, exist_ok=True)
     target = log_dir / f"{date}-{stem}.yml"
     n = 2
--- a/defendable-science/defendable_science/scaffold/layout.py
+++ b/defendable-science/defendable_science/scaffold/layout.py
@@ -13,6 +13,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+from defendable_science.core.paths import require_path_segment
+
 if TYPE_CHECKING:
     from collections.abc import Iterable, Mapping
 
@@ -142,7 +144,13 @@ class Layout:
         return self.literature_dir / "digests"
 
     def digest(self, citekey: str) -> Path:
-        """Return the digest artifact of `citekey`."""
+        """Return the digest artifact of `citekey`.
+
+        :raises LayoutError: If `citekey` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182).
+        """
+        citekey = require_path_segment(citekey, what="citekey", error=LayoutError)
         return self.digests_dir / f"{citekey}.md"
 
     # --- config ---
@@ -177,7 +185,14 @@ class Layout:
     # --- per-paper (derived, never configurable) ---
 
     def paper_dir(self, paper_id: str) -> Path:
-        """Return the root directory of `paper_id`."""
+        """Return the root directory of `paper_id`.
+
+        :raises LayoutError: If `paper_id` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182). Every other per-paper path derives from
+            this one, so guarding it here covers them all.
+        """
+        paper_id = require_path_segment(paper_id, what="paper_id", error=LayoutError)
         return self.research_root / paper_id
 
     def backlog(self, paper_id: str) -> Path:
@@ -197,7 +212,12 @@ class Layout:
         return self.paper_docs_dir(paper_id) / "positioning.md"
 
     def hypothesis_dir(self, paper_id: str, slug: str) -> Path:
-        """Return one hypothesis folder of `paper_id`."""
+        """Return one hypothesis folder of `paper_id`.
+
+        :raises LayoutError: If `paper_id` or `slug` is not a single path
+            segment (defendable-science#182).
+        """
+        slug = require_path_segment(slug, what="slug", error=LayoutError)
         return self.hypotheses_dir(paper_id) / slug
 
     def rel(self, path: Path) -> Path:
--- a/defendable-science/tests/test_layout.py
+++ b/defendable-science/tests/test_layout.py
@@ -63,6 +63,58 @@ def test_positioning_is_a_staged_paper_document() -> None:
     out = lay.Layout.default(Path("/repo"))
 
     assert lay.STAGED_DOCUMENTS[out.positioning("p1").name] == "paper"
+
+
+# --- gap 5: identifier-taking paths reject a traversal attempt (#182) -------
+
+_TRAVERSAL = "../../../../../../tmp/dsaudit/PWNED"
+
+
+def test_digest_rejects_a_traversal_citekey() -> None:
+    """A citekey is a single path segment, not a sub-path.
+
+    Reproduces #182: `Layout.digest(_TRAVERSAL).resolve()` used to land at
+    `/tmp/dsaudit/PWNED.md`, outside `repo_root` entirely.
+    """
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="citekey"):
+        out.digest(_TRAVERSAL)
+
+
+def test_paper_dir_rejects_a_traversal_paper_id() -> None:
+    """`paper_dir` is the shared root of the four derived per-paper methods.
+
+    Guarding it here covers backlog/hypotheses_dir/paper_docs_dir/positioning
+    too, since they all call through it.
+    """
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="paper_id"):
+        out.paper_dir(_TRAVERSAL)
+
+
+@pytest.mark.parametrize(
+    "method",
+    ["backlog", "hypotheses_dir", "paper_docs_dir", "positioning"],
+)
+def test_paper_dir_derived_methods_reject_a_traversal_paper_id(method: str) -> None:
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="paper_id"):
+        getattr(out, method)(_TRAVERSAL)
+
+
+def test_hypothesis_dir_rejects_a_traversal_paper_id_or_slug() -> None:
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="paper_id"):
+        out.hypothesis_dir(_TRAVERSAL, "a-real-slug")
+    with pytest.raises(lay.LayoutError, match="slug"):
+        out.hypothesis_dir("depth-collapse", _TRAVERSAL)
+
+
+def test_paper_id_containing_a_plain_slash_is_rejected_too() -> None:
+    """Not just `..` — any embedded separator would address a sub-path."""
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="paper_id"):
+        out.paper_dir("depth-collapse/../../etc")
     assert out.positioning("p1").parent == out.paper_docs_dir("p1")
 
 
--- a/defendable-science/tests/test_layout.py
+++ b/defendable-science/tests/test_layout.py
@@ -63,6 +63,7 @@ def test_positioning_is_a_staged_paper_document() -> None:
     out = lay.Layout.default(Path("/repo"))
 
     assert lay.STAGED_DOCUMENTS[out.positioning("p1").name] == "paper"
+    assert out.positioning("p1").parent == out.paper_docs_dir("p1")
 
 
 # --- gap 5: identifier-taking paths reject a traversal attempt (#182) -------
@@ -115,7 +116,6 @@ def test_paper_id_containing_a_plain_slash_is_rejected_too() -> None:
     out = lay.Layout.default(Path("/repo"))
     with pytest.raises(lay.LayoutError, match="paper_id"):
         out.paper_dir("depth-collapse/../../etc")
-    assert out.positioning("p1").parent == out.paper_docs_dir("p1")
 
 
 def test_rel_renders_a_path_for_display_and_tolerates_an_outside_path() -> None:
--- a/defendable-science/defendable_science/check/checks.py
+++ b/defendable-science/defendable_science/check/checks.py
@@ -32,6 +32,7 @@ from typing import TYPE_CHECKING, Any
 from defendable_science.check.model import Finding, Report
 from defendable_science.core.config import load_config_text
 from defendable_science.core.gitignore import literal_covers
+from defendable_science.core.paths import require_path_segment
 from defendable_science.dataset import manifest as mf
 from defendable_science.digest import artifact as artifact_mod
 from defendable_science.digest import extraction as extraction_mod
@@ -583,8 +584,9 @@ def _check_registered_paper(layout: Layout, probe: Probe, row: Row) -> list[Find
     :param row: One ``papers.md`` row.
     :returns: The findings for this row. An unbound backend is a ``gap`` —
         incomplete science in a valid file — while a root that is missing or
-        outside the repository, and a backlog the registry implies but the repo
-        does not have, are ``invalid``.
+        outside the repository, a paper-id that is a path rather than a slug,
+        and a backlog the registry implies but the repo does not have, are
+        ``invalid``.
     """
     registry = str(layout.rel(layout.papers_registry))
     paper_id = row.get("paper-id", "").strip()
@@ -600,6 +602,24 @@ def _check_registered_paper(layout: Layout, probe: Probe, row: Row) -> list[Find
                     "keys the paper across its backlog, the dashboard and `progress`"
                 ),
                 remedy="give the row a stable paper-id, or delete the row",
+            )
+        ]
+    try:
+        require_path_segment(paper_id, what="paper-id", error=ValueError)
+    except ValueError:
+        # `layout.paper_dir(paper_id)` below would raise `LayoutError` on
+        # exactly this input, uncaught — the integrity tool must report the
+        # integrity problem, not crash on it (defendable-science#182).
+        return [
+            Finding(
+                severity="invalid",
+                check=TABLES_CHECK,
+                file=registry,
+                message=f"registered paper-id {paper_id!r} is a path, not a slug",
+                remedy=(
+                    "paper-id must be a single directory name, not a path — "
+                    "e.g. `depth-collapse`, not `docs/depth-collapse`"
+                ),
             )
         ]
     if not root:
--- a/defendable-science/defendable_science/cli.py
+++ b/defendable-science/defendable_science/cli.py
@@ -2615,14 +2615,19 @@ def _positioning_context(
     :param paper: The paper id, or ``None`` to infer it from the cwd.
     :param positioning: An explicit positioning path, which always wins.
     :returns: The config mapping, the resolved layout, and the document path.
-    :raises typer.Exit: Code 1 on an invalid ``layout:`` block; code 2 when
-        ``--paper`` is omitted and the cwd is outside every paper.
+    :raises typer.Exit: Code 1 on an invalid ``layout:`` block, or a ``--paper``
+        that is not a single path segment; code 2 when ``--paper`` is omitted
+        and the cwd is outside every paper.
     """
     config, layout = _layout_or_exit()
     if positioning is not None:
         return config, layout, Path(positioning)
     paper_id = paper or _paper_dir_or_exit(layout, "--paper").name
-    return config, layout, layout.positioning(paper_id)
+    try:
+        return config, layout, layout.positioning(paper_id)
+    except LayoutError as exc:
+        typer.echo(f"invalid --paper: {exc}", err=True)
+        raise typer.Exit(code=1) from exc
 
 
 def _locator_patterns(lit: dict[str, Any] | None) -> list[re.Pattern[str]]:
@@ -2744,16 +2749,18 @@ def extract_cells(
     the same convention `extract sample --citekey` already uses.
 
     :param citekey: The digested paper's citekey (`Layout.digest`'s key).
-    :raises typer.Exit: Code 0 with the cells as JSON; code 1 if the paper has
-        not been extracted, or its cells block is absent or malformed.
+    :raises typer.Exit: Code 0 with the cells as JSON; code 1 if `citekey` is
+        not a single path segment, the paper has not been extracted, or its
+        cells block is absent or malformed.
     """
     _config, layout = _layout_or_exit()
-    artifact = layout.digest(citekey)
+    artifact: Path | None = None
     cells: list[dict[str, Any]] | None = None
     error: str | None = None
     try:
+        artifact = layout.digest(citekey)
         cells = [dataclasses.asdict(c) for c in artifact_mod.read_cells(artifact)]
-    except extraction_mod.ExtractionError as exc:
+    except (LayoutError, extraction_mod.ExtractionError) as exc:
         typer.echo(f"digest extract cells failed: {exc}", err=True)
         error = str(exc)
     typer.echo(
@@ -2761,7 +2768,7 @@ def extract_cells(
             {
                 "ok": error is None,
                 "citekey": citekey,
-                "artifact": str(artifact),
+                "artifact": str(artifact) if artifact is not None else None,
                 "cells": cells,
                 "error": error,
             },
@@ -2948,8 +2955,12 @@ def extract_record(
     errors: list[dict[str, str]] = []
     triage_not_updated: list[dict[str, str]] = []
     for citekey, paper_cells in sorted(accepted.items()):
-        artifact = layout.digest(citekey)
+        # A display-only path, computed without the guard: if `layout.digest`
+        # below rejects `citekey`, this is never opened or written — it only
+        # names, in the error report, where the write would have gone.
+        artifact = layout.digests_dir / f"{citekey}.md"
         try:
+            artifact = layout.digest(citekey)
             # `init` does not scaffold literature/digests/ — the first recorded
             # paper creates it.
             artifact.parent.mkdir(parents=True, exist_ok=True)
@@ -2961,7 +2972,7 @@ def extract_record(
                 log_dir=log_root,
                 date=date,
             )
-        except (extraction_mod.ExtractionError, OSError) as exc:
+        except (LayoutError, extraction_mod.ExtractionError, OSError) as exc:
             # The sweep continues and the report is still emitted. Aborting here
             # would leave the author knowing only that *something* failed, while
             # some papers had already landed — and re-running the whole batch to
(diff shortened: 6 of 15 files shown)


defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -197,7 +212,12 @@ def positioning(self, paper_id: str) -> Path:
         return self.paper_docs_dir(paper_id) / "positioning.md"
 
     def hypothesis_dir(self, paper_id: str, slug: str) -> Path:
-        """Return one hypothesis folder of `paper_id`."""
+        """Return one hypothesis folder of `paper_id`.
+
+        :raises LayoutError: If `paper_id` or `slug` is not a single path
+            segment (defendable-science#182).
+        """
+        slug = require_path_segment(slug, what="slug", error=LayoutError)
```
davorrunje, line 220: **This guard has no production caller.** `Layout.hypothesis_dir` is only reached from tests — grep shows no non-test call site. The code that actually creates a hypothesis folder is `exploration/backlog.py:481`, which joins the slug itself:

```python
folder = Path(paper_root) / "hypotheses" / slug
```

and `cli.py:2242` passes `--slug` straight through. Verified still exploitable on this branch: `scaffold_hypothesis(<repo>/docs/research/p1, "../../../../../../tmp/dsaudit/PWNED", …)` wrote `/tmp/tmp/dsaudit/PWNED/hypothesis.md`.

So the slug half of #182 is unclosed, and `test_hypothesis_dir_rejects_a_traversal_paper_id_or_slug` passes anyway. Route `scaffold_hypothesis` through this method (or guard at the join) and add the regression test one level up.

defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -142,7 +144,13 @@ def digests_dir(self) -> Path:
...
 
     def digest(self, citekey: str) -> Path:
-        """Return the digest artifact of `citekey`."""
+        """Return the digest artifact of `citekey`.
+
+        :raises LayoutError: If `citekey` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182).
+        """
+        citekey = require_path_segment(citekey, what="citekey", error=LayoutError)
```
davorrunje, line 153: **Split-brain with `_safe_name` on the same input.** The commit message argues for rejecting rather than coercing, and that reasoning is sound in isolation — but `literature/acquire.py:1046` still *coerces* the very same value. A `references.json` whose CSL `id` contains `/` (a DOI-shaped id from an importer, say) is happily cached by `literature acquire` as `10.1145_3292500`, while every `digest`/`extract` command on that same entry now hard-fails.

An existing repo with such a citekey therefore loses the digest half of the workflow, with no migration path and an error message that doesn't hint at one. Worth either applying the same rule at acquire time, or making this error name the coerced form the caller probably wants.

defendable-science/defendable_science/core/paths.py
```diff
@@ -0,0 +1,41 @@
...
+    :param error: The calling module's own error type (e.g. ``LayoutError``,
+        ``RecordError``) — never a bare ``ValueError`` leaking past the caller's
+        module boundary.
+    :returns: `value` unchanged.
+    :raises Exception: `error`, naming `what` and the offending value, if `value`
+        is empty, ``.``/``..``, or contains a path separator — anything that
+        would make it address more than one path segment and let it escape the
+        directory it is joined into.
+    """
+    if value in _UNSAFE or "/" in value or "\\" in value:
```
davorrunje: NUL and other control characters get through. `require_path_segment("a\u0000b", …)` passes validation and then dies at the OS layer with a bare `ValueError: embedded null byte` — a raw traceback of exactly the kind this function exists to convert into the caller's own error type.

```python
if value in _UNSAFE or "/" in value or "\\" in value or any(c < " " for c in value):
```

(or an explicit `"\x00" in value` if you'd rather keep the rule narrow to what actually breaks).

defendable-science/tests/test_layout.py
```diff
@@ -63,6 +63,58 @@ def test_positioning_is_a_staged_paper_document() -> None:
...
+    with pytest.raises(lay.LayoutError, match="slug"):
+        out.hypothesis_dir("depth-collapse", _TRAVERSAL)
+
+
+def test_paper_id_containing_a_plain_slash_is_rejected_too() -> None:
+    """Not just `..` — any embedded separator would address a sub-path."""
+    out = lay.Layout.default(Path("/repo"))
+    with pytest.raises(lay.LayoutError, match="paper_id"):
+        out.paper_dir("depth-collapse/../../etc")
     assert out.positioning("p1").parent == out.paper_docs_dir("p1")
```
davorrunje: This assertion belongs to `test_positioning_is_a_staged_paper_document` — the new gap-5 block was pasted in above it, orphaning it into the tail of the traversal test. It still executes so nothing fails, but it no longer lives in the test that documents it, and a later edit to the traversal test could drop it unnoticed. Move it back up to line 65.

davorrunje (commented):
## Code review — `/code-review PR #183`

The guard itself is correct and the reject-over-coerce rationale in 54dcdce's message is right. The problems are **placement and integration**: the one method that actually needed the slug guard has no production caller while the real writer bypasses it, and the new `LayoutError` is unhandled at every real call site — including inside `check`, whose job is to report bad content rather than crash on it, and inside three loops whose own comments promise the sweep continues.

Four findings are left as inline comments. The five below are in files this PR does not touch, so they have no diff line to anchor to.

---

### 1. `check` now crashes instead of reporting — `check/checks.py:614` (high)

`_check_registered_paper` builds its **remedy string** with `layout.rel(layout.paper_dir(paper_id))`, where `paper_id` is `row.get("paper-id")` straight out of human-authored `papers.md`. A row with `paper-id: ../evil` and an empty `root` now raises an uncaught `LayoutError` where it previously produced a clean `invalid` Finding. Same at `checks.py:635` (the outside-repo branch). Neither `run_checks` nor `cli.py:641` has a handler.

Reproduced: `LayoutError: paper_id is not a valid path segment: '../evil'` tracebacks out of `checks.py:614`.

`check` is the command that exists to *report* malformed content. It must not crash on it — build the remedy from the raw string, or catch `LayoutError` and emit the `invalid` Finding.

### 2. The hypothesis-slug traversal is still open — `exploration/backlog.py:481` (high)

`Layout.hypothesis_dir` is guarded, but per grep it has **zero production callers** (tests only), so that half of the change is dead code. The real writer, `scaffold_hypothesis`, does its own join:

```python
folder = Path(paper_root) / "hypotheses" / slug
```

and `cli.py:2242` passes `--slug` through unchecked.

Reproduced: `scaffold_hypothesis(<repo>/docs/research/p1, "../../../../../../tmp/dsaudit/PWNED", …)` wrote `/tmp/tmp/dsaudit/PWNED/hypothesis.md`. `backlog promote --scaffold --level hypothesis --slug ../../…` still escapes the repo.

Route `scaffold_hypothesis` through `layout.hypothesis_dir` (or guard the slug where it joins), and add a regression test at that level — the existing `test_hypothesis_dir_rejects_a_traversal_paper_id_or_slug` passes while the hole is open.

### 3. Raw tracebacks at every CLI call site — `cli.py:2751`, `2625`, `2260`, `3252`, `3590` (high)

No call site catches `LayoutError`; `_layout_or_exit` wraps only `resolve_layout`, not the identifier methods. Against the repo's *failure honesty* rule ("never let a transient error surface as a raw traceback").

Reproduced: `defendable-science digest extract cells --citekey '../../../../tmp/PWNED'` prints a full Rich traceback ending in `LayoutError: citekey is not a valid path segment`.

### 4. Guard placed outside the `try` in three sweep loops — `cli.py:2951`, `3103`, `3572-3590` (medium)

`artifact = layout.digest(citekey)` sits **above** the `try` whose comment states the sweep must continue:

> Aborting here would leave the author knowing only that *something* failed, while some papers had already landed.

Citekeys here come from the `--cells` JSON payload and `extraction_mod.validate` does no path-segment check, so one entry with `citekey: "../x"` now aborts mid-sweep with a traceback *after* earlier papers were written and their log entries appended — the exact failure the comment forbids. Move the call inside the `try`.

### 5. Same placement bug in `_apply_verdict` — `cli.py:3302`, `3252` (medium)

Docstring: *"A failure on one member is reported and the sweep continues; aborting would leave the batch half-marked with no report of which half."* `target = layout.digest(key)` is above the `try`, so a bad member citekey does exactly that.

### 6. Stale `:raises:` contract on `scaffold_paper` — `exploration/backlog.py:598` (medium)

`scaffold_paper` can now raise `LayoutError` from `layout.paper_dir(paper_id)`, but its docstring still declares only `:raises BacklogError:` and `cli.py:2339` catches only `backlog_mod.BacklogError`. `paper_id` is `row["id"]` from the portfolio-backlog table (external input), so a row with a slash in its id tracebacks instead of exiting 1.

---

**Suggested split for a follow-up session:** #1 + #2 are the substantive ones (a crash in the integrity checker, and the traversal the PR was opened to close). #3–#6 are a single consistent pass over `cli.py`/`backlog.py` error handling. The inline comments cover the rest.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed and fixed in a6f8063 — reproduced the exact exploit first: `scaffold_hypothesis(<root>, "../../../../../../tmp/dsaudit/PWNED", ...)` did write outside the repo on this branch. Guarded the join in `scaffold_hypothesis` itself with the same `require_path_segment` used everywhere else in this PR, rather than trying to route it back through `Layout.hypothesis_dir` — `scaffold_hypothesis` takes a bare `paper_root: str | Path`, not a `Layout`, so there's no natural way to call through that method without a larger signature change outside this PR's scope. New regression test reproduces your exact exploit and confirms it raises `BacklogError` naming `slug`.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed the coercion — `_safe_name("10.1145/3292500")` does produce `10.1145_3292500`, and I agree this is a real, live inconsistency for a DOI-shaped citekey.

Not fixing `acquire.py`'s behavior in this PR, though. Changing an established, already-tested coercion policy on a different module's boundary is a bigger decision than "guard the traversal in `Layout`" — it needs its own call on which policy (reject vs. coerce) is canonical, and ideally where the citekey should be validated once rather than at every downstream consumer. Filed #184 with your exact reproduction and both remedies you named (reject at acquire time, or make the rejection message actionable) as open options, not a decision.

One thing worth noting since you flagged "no migration path": this isn't a regression #183 introduced in the sense of a new failure mode. Pre-#183, the same citekey would have silently created a *nested* directory (`digests/10.1145/3292500.md") via unvalidated `Path` joining — already diverging from `acquire`'s flat mangled cache name, just silently instead of loudly. #183 turned a silent wrong-nesting into a loud rejection; it didn't create the underlying two-policies problem, but its own review is what surfaced it.

defendable-science/defendable_science/core/paths.py
davorrunje: Confirmed — `require_path_segment("a\x00b", ...)` passed validation on this branch and would have died downstream as a bare `ValueError: embedded null byte`, exactly the failure-honesty violation this function exists to prevent.

Fixed in f924140 with your suggested `any(c < " " for c in value)`, catching NUL and every other control character in one check rather than special-casing NUL alone. Added NUL, newline, tab and a bare 0x1f to the parametrized rejection test.

defendable-science/tests/test_layout.py
davorrunje: Confirmed — that assertion had nothing to do with the traversal test it was trailing. Moved in 989d1e3 into `test_positioning_is_a_staged_paper_document`, which already exercises the same paper_id ("p1") and the same concern (positioning's relationship to paper_docs_dir), rather than back to line 65 literally — that's where the STAGED_DOCUMENTS assertion already lives, and this one is a natural addition to it rather than a separate line.

defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -177,7 +185,14 @@ def kappa_dir(self) -> Path:
...
     def paper_dir(self, paper_id: str) -> Path:
-        """Return the root directory of `paper_id`."""
+        """Return the root directory of `paper_id`.
+
+        :raises LayoutError: If `paper_id` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182). Every other per-paper path derives from
+            this one, so guarding it here covers them all.
+        """
+        paper_id = require_path_segment(paper_id, what="paper_id", error=LayoutError)
```
davorrunje, line 195: **`check` now tracebacks on the exact malformation it exists to report.** (Real site: `defendable-science/defendable_science/check/checks.py:614` and `:635`.)

`_check_registered_paper` takes `paper_id` verbatim from a human-authored `papers.md` row (`row.get("paper-id", "").strip()`, line 590) and interpolates `layout.rel(layout.paper_dir(paper_id))` into the *remedy string* of the finding it is about to return. With this guard, a row whose paper-id is not a bare segment raises `LayoutError` from inside the validator — and nothing up the stack catches it (`checks.py:2114`'s `except LayoutError` wraps `resolve_layout` in the config check, a different function).

Reproduced on this branch:

```python
>>> from defendable_science.check import checks
>>> from defendable_science.scaffold.layout import Layout
>>> lay = Layout.default(Path('/repo'))
>>> checks._check_registered_paper(lay, probe, {'paper-id': 'docs/dc', 'root': ''})
LayoutError: paper_id is not a valid path segment: 'docs/dc'
```

So a researcher who types the *path* into the paper-id column — a plausible mistake, and precisely what `TABLES_CHECK` is there to catch — gets a Python traceback out of `defendable-science check` instead of the `invalid` Finding with its remedy. The integrity tool refuses to report the integrity problem.

Both sites build the remedy on a `paper_id` that is by definition suspect (line 592 already rejected the empty case). The remedy string wants the *intended* path, so the natural fix is to render it without going through the guarded method, or to catch `LayoutError` around the row and emit an `invalid` Finding naming the unusable paper-id — which is the more useful output anyway.

defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -177,7 +185,14 @@ def kappa_dir(self) -> Path:
...
         return self.research_root / paper_id
```
davorrunje, line 196: **`scaffold_paper` leaks `LayoutError` past the CLI's `BacklogError` handler** — the paper-level sibling of the hypothesis-level bug a6f8063 just fixed. (Real site: `exploration/backlog.py:602`, handler at `cli.py:2339`.)

`scaffold_paper` documents `:raises BacklogError: If the paper root already exists.` and now also raises `LayoutError` via `layout.paper_dir(paper_id)` at line 602. `backlog promote` catches only `backlog_mod.BacklogError`:

```python
    except backlog_mod.BacklogError as exc:      # cli.py:2339
        typer.echo(str(exc), err=True)
        raise typer.Exit(code=1) from exc
```

and `paper_id` is `row["id"]` (`cli.py:2251`), read straight from the human-authored portfolio backlog. Reproduced:

```python
>>> b.scaffold_paper(lay, '../../../../tmp/PWNED', 'x', backend='y')
LayoutError: paper_id is not a valid path segment: '../../../../tmp/PWNED'   # uncaught
```

`backlog promote --scaffold --level paper` therefore exits on a raw traceback rather than the documented "code 1 on a guard violation" (docstring at `cli.py:2318`). Note the asymmetry this PR now carries: `scaffold_hypothesis` raises `BacklogError`, which *is* caught; its paper-level twin raises `LayoutError`, which is not. Easiest fix is to translate at `backlog.py:602` (`except LayoutError as exc: raise BacklogError(str(exc)) from exc`), which also makes the docstring true again.

defendable-science/defendable_science/scaffold/layout.py
davorrunje, line 153: **`digest extract record` now aborts mid-sweep on a bad citekey — the one outcome its own comment forbids.** (Real site: `cli.py:2951`.)

```python
    for citekey, paper_cells in sorted(accepted.items()):
        artifact = layout.digest(citekey)          # 2951 -- OUTSIDE the try
        try:
            artifact.parent.mkdir(parents=True, exist_ok=True)
            log_entry = artifact_mod.write_extraction(...)
        except (extraction_mod.ExtractionError, OSError) as exc:
            # The sweep continues and the report is still emitted. Aborting here
            # would leave the author knowing only that *something* failed, while
            # some papers had already landed -- and re-running the whole batch to
            # find out would append a second log entry for each of those.
```

`accepted` is keyed by `Cell.citekey`, parsed from the cells JSON on stdin; nothing in `digest/extraction.py` constrains a citekey to a path segment (`validate` only checks the axis, the locator and the `SELF_ROW` self-reference). So a cells payload with `"citekey": "10.1145/3292500"` — or any key with a separator — raises `LayoutError` at 2951, *before* the `try`, and `LayoutError` isn't in the `except` tuple either way.

Concretely, with `accepted = {"a-good": [...], "b/bad": [...]}`: `a-good`'s artifact is written and its log entry appended, then the command dies with a traceback. No report is emitted, so the author cannot tell which half landed — and the re-run appends a second log entry for `a-good`, which is exactly what the comment says must not happen. Move `layout.digest(citekey)` inside the `try` and add `LayoutError` to the caught tuple.

Same shape at `cli.py:3103` (`digest depth cells record`, whose comment makes the identical promise).

davorrunje (commented):
Round 2. The three code fixes from round 1 all check out (`scaffold_hypothesis` guard reproduces and rejects; `any(c < " " for c in value)` catches NUL/CR/LF/0x1f; the orphaned assertion is back where it belongs), and #184 is a fair place for the `_safe_name` policy question. Resolved all four threads.

What this round found is a single systemic consequence the round-1 pass and I both missed: **`Layout.paper_dir` / `Layout.digest` were total functions and are now partial, but none of their nine call sites learned to catch `LayoutError`.** The guard is right; the blast radius isn't contained. `uv run pytest -q` is green at 100% coverage because no test drives a bad identifier through a *caller* — only through `Layout` directly.

The consequences fall in three tiers, worst first:

1. `check`'s registry validator now dies on the exact malformation it exists to report (reproduced below).
2. Three `digest` sweeps whose docstrings *explicitly promise* "a failure on one member is reported and the sweep continues" now abort mid-batch with a raw traceback, after earlier members have already been written and their append-only log entries appended.
3. `scaffold_paper` and three single-shot commands surface a raw traceback where the docstring promises a typed error or a JSON report.

All of it is the repo's failure-honesty rule ("never let a transient error surface as a raw traceback") — and it's the same class of inconsistency this PR was written to fix, one level up: the rule got applied at the definition and not at the readers.

Seven comments below. Several are anchored on the nearest diff line rather than the offending line, since GitHub won't take a comment outside the hunk — each names its real `file:line`.

defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -142,7 +144,13 @@ def digests_dir(self) -> Path:
...
     def digest(self, citekey: str) -> Path:
-        """Return the digest artifact of `citekey`."""
+        """Return the digest artifact of `citekey`.
+
+        :raises LayoutError: If `citekey` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182).
+        """
+        citekey = require_path_segment(citekey, what="citekey", error=LayoutError)
         return self.digests_dir / f"{citekey}.md"
```
davorrunje, line 154: **`extract sample --apply` can leave the batch half-marked with no report of which half.** (Real sites: `cli.py:3302` in `_apply_verdict`, and `cli.py:3252` in `_read_sampled_cells`.)

Same `layout.digest(key)`-outside-the-`try` shape as the previous comment, but the consequence is worse because `_apply_verdict` writes to *every* member and appends an accountability-log entry per sampled one:

```python
    for key in members:
        target = layout.digest(key)                # 3302 -- outside the try
        try:
            artifact_mod.set_batch_check(target, verdict, date=date)
            ...
            artifact_mod.append_check_log(...)
        except (...):                              # LayoutError not caught
```

Its docstring is explicit: *"A failure on one member is reported and the sweep continues; aborting would leave the batch half-marked with no report of which half."* One member whose citekey isn't a bare segment now produces precisely that: members earlier in the list already have `status.batch-check` written and check-log entries appended, then the command tracebacks and the `errors`/`updated`/`log_entries` report is never emitted. For an append-only evidence trail that is the expensive failure — the re-run duplicates the log entries of everything that already landed.

`_read_sampled_cells` (3252) has the same structure and the same promise (*"A paper whose cells cannot be read is an error, never an empty cell list"*), as does `_render_rows` at `cli.py:3590` (*"A paper whose artifact cannot be read contributes no row rather than an empty one"*). All three need `layout.digest(...)` inside the `try` with `LayoutError` in the caught tuple, so a bad key becomes one `errors` entry and the sweep goes on.

defendable-science/defendable_science/scaffold/layout.py
```diff
@@ -142,7 +144,13 @@ def digests_dir(self) -> Path:
         return self.literature_dir / "digests"
 
     def digest(self, citekey: str) -> Path:
-        """Return the digest artifact of `citekey`."""
+        """Return the digest artifact of `citekey`.
+
+        :raises LayoutError: If `citekey` is not a single path segment — see
+            :func:`~defendable_science.core.paths.require_path_segment`
+            (defendable-science#182).
+        """
```
davorrunje, line 152: **Two single-shot commands turn a bad `--citekey`/`--paper` into a traceback instead of their documented failure signal.**

`cli.py:2751` — `digest extract cells`:

```python
    artifact = layout.digest(citekey)     # 2751 -- before the try/except
    cells = None
    error = None
    try:
        cells = [... artifact_mod.read_cells(artifact) ...]
    except extraction_mod.ExtractionError as exc:
        ...
```

The command's contract (docstring at 2747) is a JSON report plus `code 1` on failure; the whole point of the `error` variable is that a failure is *reported*, not raised. `digest extract cells --citekey ../x` now dies on an uncaught `LayoutError` before the report structure exists.

`cli.py:2625` — `_resolve_positioning`: `layout.positioning(paper_id)` where `paper_id` is the `--paper` option (or a cwd-derived name). Docstring promises `typer.Exit` code 1 or 2; `--paper ../x` gets a traceback. This one feeds `extract axes`, `extract record` and `depth cells record`, so it is on the common path.

There is no global exception handler in `cli.py` (`main` at 195 installs none, and Typer's default is to print the traceback), so nothing catches these further out. Both want the treatment the module already gives `LayoutError` at `cli.py:323`/`435`: catch, echo, `typer.Exit`.

defendable-science/defendable_science/defend/record.py
```diff
@@ -338,7 +339,11 @@ def append_log_entry(log_dir: Path, date: str, stem: str, body: str) -> Path:
         naming scheme changes.
     :param body: The rendered YAML to write.
     :returns: The path written.
+    :raises RecordError: If `stem` is not a single path segment — see
+        :func:`~defendable_science.core.paths.require_path_segment`
+        (defendable-science#182).
     """
+    stem = require_path_segment(stem, what="stem", error=RecordError)
     log_dir.mkdir(parents=True, exist_ok=True)
     target = log_dir / f"{date}-{stem}.yml"
```
davorrunje, line 351: The guard covers `stem` but not `date`, and both are interpolated into the same filename:

```python
    stem = require_path_segment(stem, what="stem", error=RecordError)
    log_dir.mkdir(parents=True, exist_ok=True)
    target = log_dir / f"{date}-{stem}.yml"
```

Being straight: **no current caller misbehaves.** All four (`_append_log`, and `digest/artifact.py:627`/`735`/`930`) pass `date_cls.today().isoformat()` or `today_iso()`, so `date` is machine-generated today. Flagging it because `append_log_entry` is public API, its docstring now advertises a validated-input contract that covers only half the name it builds, and `:param date:` says merely "ISO date" without enforcing it. A future caller wiring a `--date` option through (the pattern already exists — `backlog promote --date` at `cli.py:2291`) reopens #182 at a boundary that now *looks* guarded. One more `require_path_segment(date, what="date", error=RecordError)` closes it for good; low priority, your call.

defendable-science/defendable_science/core/paths.py
```diff
@@ -0,0 +1,42 @@
...
+        ``RecordError``) — never a bare ``ValueError`` leaking past the caller's
+        module boundary.
+    :returns: `value` unchanged.
+    :raises Exception: `error`, naming `what` and the offending value, if `value`
+        is empty, ``.``/``..``, contains a path separator, or contains a control
+        character (e.g. an embedded NUL) - anything that would either address
+        more than one path segment or reach the OS call as a raw, uncaught
+        error instead of this function's own signal.
+    """
+    if value in _UNSAFE or "/" in value or "\\" in value or any(c < " " for c in value):
```
davorrunje: A Windows drive-relative value passes and still escapes. The explicit `"\\" in value` check says Windows is in scope for this guard, but `"C:PWNED"` contains no separator, isn't in `_UNSAFE`, and has no control characters — so it is accepted, and the join discards the parent entirely:

```python
>>> from pathlib import PureWindowsPath as W
>>> W('/repo/docs/research') / 'C:PWNED'
PureWindowsPath('C:PWNED')
```

Same for a bare `"C:"`. On a Windows checkout that lands the artifact in the current directory of drive `C:`, wherever that happens to be — outside `repo_root`, which is what #182 is about. (`"C:../../x"` is already caught by the `/` check, so it is specifically the separator-free drive-relative form that slips through.)

Low severity: CI is `ubuntu-latest` only and there is no `Operating System` classifier in `pyproject.toml`, so Windows is not a supported target today. But the `\\` check already pays the cost of caring about Windows, and `os.path.splitdrive(value)[0]` (or `PureWindowsPath(value).drive`) finishes the job in one line. Worth it while the guard is fresh — or worth dropping the `\\` check and documenting POSIX-only. The current state is halfway.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed and fixed in cc7ff42: reproduced your exact repro first (`LayoutError: paper_id is not a valid path segment: 'docs/dc'` from `_check_registered_paper`). Added an explicit `require_path_segment` check right after the existing empty-paper-id branch, returning an `invalid` Finding naming the bad paper-id rather than letting the later `layout.paper_dir(paper_id)` remedy-building calls raise. Regression test: `test_tables_check_flags_a_paper_id_that_is_a_path_instead_of_a_slug`.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed and fixed in 4be3f33 — reproduced your exact repro (`b.scaffold_paper(lay, '../../../../tmp/PWNED', 'x', backend='y')` raising uncaught `LayoutError`). Wrapped the one `layout.paper_dir(paper_id)` call and translated to `BacklogError`, the type `backlog promote`'s handler at `cli.py:2339` actually catches — same treatment the hypothesis-level sibling got a round earlier. Regression test: `test_scaffold_paper_rejects_a_traversal_paper_id`.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed and fixed in 46f6d52. Both `digest extract record` and `digest depth cells record` now catch `LayoutError` in the per-citekey sweep's except tuple, with `layout.digest(citekey)` moved inside the try. A display-only path (`layout.digests_dir / f"{citekey}.md"`, never opened) names the failure in the error entry when `digest()` itself is what raised — the existing error-dict shape didn't need to change.

Two regression tests (one per command, since you named both): `test_record_reports_a_traversal_citekey_and_still_records_the_rest` and its depth-cells sibling. Both confirm the good citekey still lands and its artifact exists on disk, and both assert `result.exception is None or isinstance(result.exception, SystemExit)` rather than checking stdout/stderr for "Traceback" — `CliRunner`'s default `catch_exceptions=True` swallows any exception into `exit_code == 1` without ever writing a traceback to either stream, so that check would have passed either way. Confirmed both tests genuinely fail against the pre-fix code via the real discriminator, `type(result.exception)`.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Confirmed and fixed in 46f6d52, same commit — all three: `_apply_verdict`, `_read_sampled_cells`, and `_cells_to_render` (the actual name of the function at the `render` call site; matches what you described).

For `extract sample --apply`, one test covers both functions in a single run: a 2-member batch (one good citekey, one traversal) is drawn whole (`select_sample`'s size floor — `min(2, max(3,1))=2`), so `_read_sampled_cells` hits the bug first, then `_apply_verdict` hits it again for the same bad key. Verified `_note_error`'s existing dedup ("reporting it twice would suggest two problems") correctly collapses both hits into a single `errors` entry rather than two — asserted explicitly, not just checked for presence.

One correction to my own test along the way: I initially tried `--verdict verified`, which doesn't reach `_apply_verdict` at all for this case — the command has a separate, deliberate refusal ("refusing to record a verified batch — N of the drawn papers could not be shown to the human") that short-circuits before it. Switched to `--verdict failed`, matching the existing precedent (`test_a_sampled_paper_whose_cells_could_not_be_read_is_not_marked_checked`) for exercising the sweep-continuation path specifically.

defendable-science/defendable_science/scaffold/layout.py
davorrunje: Both confirmed and fixed in 46f6d52.

`digest extract cells` (2751): moved `layout.digest(citekey)` inside the existing try, added `LayoutError` to the except tuple, and it's reported through the same JSON envelope as an `ExtractionError` — `artifact` renders `null` when it couldn't even be computed (pinned by an added assertion; the happy-path assertion already existed for the non-null case). Regression test: `test_cells_a_traversal_citekey_is_reported_not_raised`.

`_positioning_context` (2625): wrapped `layout.positioning(paper_id)` in its own try/except, echoing and raising `typer.Exit(code=1)` — matching the module's existing `LayoutError`-at-a-boundary idiom at `cli.py:323`/`435` you pointed to. Since all four callers route through this one function, one fix covers `extract axes`, `extract record`, `depth cells record` and `render` for the `--paper` path specifically. Regression test exercises it via `extract axes --paper ../x`.

Both tests assert `result.exception is None or isinstance(result.exception, SystemExit)` rather than checking for "Traceback" in stdout/stderr, for the same `CliRunner` reason noted on the sibling comment.

defendable-science/defendable_science/defend/record.py
davorrunje: Taken — cheap and you're right that it closes the guarantee for the whole filename rather than half of it. Fixed in a788813: `append_log_entry` now validates `date` the same way as `stem`, with a docstring note that no current caller passes an untrusted value but the function is public API. Regression test: `test_append_log_entry_rejects_a_traversal_date`.

defendable-science/defendable_science/core/paths.py
davorrunje: Confirmed exactly — `"C:PWNED"` has no separator and isn't in the unsafe set, so it passed every existing check, and `PureWindowsPath` does discard the parent on join.

Took the "finish the job" option rather than dropping the backslash check, fixed in a788813. Used `ntpath.splitdrive` specifically rather than `os.path.splitdrive` — the latter is a no-op for drive detection on POSIX, which is what this suite's CI actually runs on, so an `os.path`-based check would pass in every test run regardless of whether it worked. Added `"C:PWNED"`, `"C:"` and `"c:PWNED"` to the existing parametrized rejection test, and confirmed all three fail without the fix.

davorrunje:
Round 3 addressed and pushed — 4 commits (`cc7ff42`, `4be3f33`, `46f6d52`, `a788813`).

**The theme this round:** `require_path_segment` gave `Layout.digest`/`paper_dir`/`positioning` a failure mode none of their existing callers were written to expect, and this round found every place that broke — 9 findings total.

| Site | Fix |
|---|---|
| `check/checks.py` (`_check_registered_paper`) | Validates paper-id explicitly, returns an `invalid` Finding instead of letting the check itself crash |
| `backlog.py` (`scaffold_paper`) | Translates `LayoutError` → `BacklogError`, the type the CLI handler actually catches |
| `digest extract cells` | `LayoutError` joins the existing except tuple; reported through the same JSON envelope |
| `_positioning_context` (shared by `extract axes`/`record`, `depth cells record`, `render`) | Own try/except, `typer.Exit(1)` — matches the module's existing pattern |
| `digest extract record`, `digest depth cells record` (sweeps) | `layout.digest` moved inside the per-citekey try; a bad citekey no longer aborts a batch mid-way |
| `_read_sampled_cells` + `_apply_verdict` (`extract sample --apply`) | Same fix in both; verified the existing error-dedup collapses a citekey that fails in both into one report entry, not two |
| `_cells_to_render` (`extract render`) | Same shape, same fix |
| `core/models.py`... `core/paths.py` (`require_path_segment`) | Now rejects control characters (was reaching the OS as a raw `ValueError`) and Windows drive-relative values (`"C:PWNED"`) |
| `defend/record.py` (`append_log_entry`) | `date` gets the same guard as `stem` — no live exploit, but the function is public API and was only half-guarded |

**One thing worth flagging on my own review process:** several of my new tests initially asserted `"Traceback" not in (result.stdout + result.stderr)` to prove an exception was handled cleanly. That check is meaningless under `CliRunner`'s default `catch_exceptions=True` — it swallows *any* exception into `exit_code == 1` without ever writing a traceback to either stream, so the assertion would pass whether the bug was fixed or not. Caught it by checking `result.exception`'s actual type (`SystemExit` for a deliberate `typer.Exit`, the raw exception otherwise) instead, and re-verified every new CLI test in this round fails against pre-fix code via that real discriminator, not the string check.

1837 tests, 100% statement+branch coverage, mypy strict and ruff clean.

PR #185, opened 2026-08-29
test(literature): pin both halves of the malformed-body reason string (closes #178)

### LINKED ISSUE #178: test: pin the reason string in resolve()'s transport-error assertions
## Context

#169 established a house rule for boundary-error tests, after four separate fix rounds spent on
it: an assertion on a validation failure must pin **both** the source (the file or URL that
failed) and the reason (why). Either half alone lets a regression through — a message that
silently stopped naming its source would keep passing a reason-only assertion, and ADR-0043's
decision point 4 makes "a failure names its source and its reason" the guarantee the whole
parsing-boundary design exists to deliver.

One test added late in that branch does not follow the rule:

```python
# defendable-science/tests/test_literature.py
def test_resolve_s2_crossref_malformed_body_is_transport_error_not_a_miss() -> None:
    client = _client({f"{_S2}/paper/CorpusId:7": "not-a-dict"})
    rec = graph.resolve("CorpusId:7", client=client)
    assert rec["resolved"] is False
    assert rec["transport_error"] is True
```

`resolve()` also populates `rec["reason"]`, which currently reads:

```
https://api.semanticscholar.org/graph/v1/paper/CorpusId:7: <root>: Input should be a valid dictionary or instance of S2ExternalIds (got str)
```

Nothing asserts that. A regression that emptied or truncated `reason` — dropping the URL, or
the field path — would leave this test green, and `reason` is what a user actually reads when
`literature resolve` exits 3.

This was found by #169's final whole-branch review and deliberately parked rather than fixed:
the process allows no second fix wave after a final review, the test does pin the discriminator
that was actually broken (exit 1 "genuine miss" versus exit 3 `transport_error`), and the
`_explain` code path that builds the message is source-pinned by assertions in
`tests/test_models.py`, `tests/test_keys.py`, `tests/test_acquire.py` and
`tests/test_manifest.py`. So the guarantee is covered globally; it just is not localised to
`resolve()`.

## Goal

The `resolve()` S2 transport-error test pins the reason string as well as the discriminator, so
a regression in that specific message is caught where it happens rather than only by a sibling
suite.

## Where

- `defendable-science/tests/test_literature.py` —
  `test_resolve_s2_crossref_malformed_body_is_transport_error_not_a_miss`
- `defendable-science/defendable_science/literature/graph.py` — `resolve` and its
  `_resolve_s2_xref` helper, which produce the `reason` value (no production change expected)

## Proposed approach

Add an assertion requiring `rec["reason"]` to contain both the failing S2 URL and the failure
reason, in the idiom the rest of the suite settled on:

```python
    assert re.search(
        rf"{re.escape(f'{_S2}/paper/CorpusId:7')}.*valid dictionary", rec["reason"]
    )
```

Read the actual runtime message before writing the pattern rather than copying the one quoted
above — it will drift as Pydantic's wording changes, and the point is to pin what the code
really emits.

While there: check whether `resolve()`'s **OpenAlex** malformed-200 test
(`test_resolve_malformed_200_body_is_transport_error_not_a_miss`) has the same gap. If it does,
fix both in one change — they are the same assertion on the same field of the same function.

Test-only. No production change; the messages are already correct.

## Acceptance criteria

- [ ] The S2 test asserts `rec["reason"]` names both the failing URL and the reason
- [ ] The OpenAlex sibling test is checked and, if it has the same gap, fixed the same way
- [ ] Each assertion would fail if the URL were dropped from the message — verify by
      temporarily mutating the message, not by inspection alone

## References

- #169 — the branch that added the test; its final whole-branch review parked this
- `decisions/0043-pydantic-at-the-parsing-boundary.md` — decision point 4 (a failure surfaces as
  an explicit signal carrying the field path and reason)
- `defendable-science/defendable_science/core/models.py` — `_explain`, which builds the message
- `defendable-science/tests/test_acquire.py` and `tests/test_keys.py` — the established
  both-halves assertion idiom to copy

davorrunje:
## What

Pins the full validation-failure message — both the failing URL and the
reason text — on the two `resolve()` transport-error regression tests
(OpenAlex and S2 crossref legs), closing a gap where a regression that
emptied or truncated `rec["reason"]` could have stayed green.

## Details

- `test_resolve_malformed_200_body_is_transport_error_not_a_miss` and
  `test_resolve_s2_crossref_malformed_body_is_transport_error_not_a_miss`
  previously asserted only the `transport_error` discriminator.
- Added `re.search` assertions against `rec["reason"]` following the same
  idiom already used in `tests/test_keys.py`.
- Verified each assertion is load-bearing: temporarily dropped the URL
  from `_explain`'s output in `core/models.py`, confirmed both tests fail,
  then reverted — no production code change ships in this PR.
- `uv run pytest -q` (100% statement+branch coverage gate), `uv run mypy`,
  `uv run ruff check`, and `uv run ruff format --check` all pass.

Closes #178

### DIFF
--- a/defendable-science/tests/test_literature.py
+++ b/defendable-science/tests/test_literature.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import re
 from typing import Any
 
 import pytest
@@ -1226,6 +1227,8 @@ def test_resolve_malformed_200_body_is_transport_error_not_a_miss() -> None:
     rec = graph.resolve("W1", client=client)
     assert rec["resolved"] is False
     assert rec["transport_error"] is True
+    pattern = rf"{re.escape('https://api.openalex.org/works/W1')}: <root>: Input should be a valid dictionary"
+    assert re.search(pattern, rec["reason"])
 
 
 # --- S2 leg fix wave: converted halfway, whole-branch review (#169) ---------
@@ -1305,6 +1308,8 @@ def test_resolve_s2_crossref_malformed_body_is_transport_error_not_a_miss() -> N
     rec = graph.resolve("CorpusId:7", client=client)
     assert rec["resolved"] is False
     assert rec["transport_error"] is True
+    pattern = rf"{re.escape(f'{_S2}/paper/CorpusId:7')}: <root>: Input should be a valid dictionary"
+    assert re.search(pattern, rec["reason"])
 
 
 # --- final-review fix wave: dropped null guards (default_factory covers a


davorrunje:
## Code review: one low-severity finding — over-pinned to Pydantic's wording

Reviewed at e1510aa. The change does what #178 asked (both halves of the `reason` are now pinned, and the assertions are not vacuous — I confirmed the URLs match the fake client's keys and the `<root>:` prefix matches `_explain()` at `defendable-science/defendable_science/core/models.py:69`). Full suite: 1796 passed, 28 skipped, coverage 100.00%.

Not blocking a merge decision, but worth fixing — **leaving this to a follow-up session rather than fixing it here.**

### Finding: the assertions pin a vendor-owned string

`defendable-science/tests/test_literature.py:1230` and `:1313` assert Pydantic's own message text:

```python
pattern = rf"{re.escape('https://api.openalex.org/works/W1')}: <root>: Input should be a valid dictionary"
assert re.search(pattern, rec["reason"])
```

`"Input should be a valid dictionary"` is Pydantic's `dict_type`/`model_type` message, and `defendable-science/pyproject.toml:30` allows `pydantic>=2.13,<3`. Any minor Pydantic release that rewords it breaks both tests while `graph.py` is still behaving correctly — a vendor-wording change presenting as a regression.

The neighbouring `test_enrich_work_rejects_a_string_publication_year` gets this right: it matches only the field path, so it is version-stable.

**Suggested fix** — pin the parts this repo owns (the `{source}: ` prefix and the `<root>` location), leave the vendor sentence unpinned. That still satisfies #178's "both halves".

### Secondary: the regex machinery is unnecessary

Both patterns are `re.escape`d literals, so `re.search` is doing a plain substring test. `assert "<literal>" in rec["reason"]` is equivalent, matches the idiom already used for `reason` assertions elsewhere in this file (lines 393, 634, 1328), and drops the new `import re`. A substring `in` check tolerates the S2 message's longer tail (`… or instance of S2ExternalIds (got str)`) exactly as `re.search` does.

### Acceptance criteria for the follow-up

- [ ] Both assertions pin `f"{url}: <root>: "` (URL + location) and no vendor message text.
- [ ] `import re` removed from `defendable-science/tests/test_literature.py` if nothing else uses it.
- [ ] Mutation check still holds: dropping the `{source}: ` prefix in `parse_obj`, or dropping `err['msg']` from `_explain`, must still fail these tests. (If dropping the vendor text means the `err['msg']` mutation no longer fails, assert non-emptiness of the tail rather than its content.)
- [ ] `uv run pytest -q` passes with the 100% branch-coverage gate intact.

PR #186, opened 2026-08-29
fix(literature,dataset): stop two silent-wrong-value boundaries

### LINKED ISSUE #173: literature+dataset: eliminate the silent-wrong-value boundaries #169 does not cover (OpenAlex filter injection, manifest `str()` coercion)
## Context

The codebase audit in #172 (`.codebase_audit/be-tech-debt.md`, BE-4 and BE-9) drew a severity distinction this repo should adopt explicitly: **a boundary that produces a plausible wrong value is worse than one that raises.** A traceback is ugly but self-announcing; a wrong value that looks right enters the research record and is never questioned. For a tool whose output is meant to be defensible, the second is the failure mode that matters, and the repo's own failure-honesty rule (`CLAUDE.md`) is written against exactly it.

#169 covers six unvalidated **JSON** boundaries, of which two are silent-wrong-value (`literature/graph.py:348` and `:110`). **This issue covers the two silent-wrong-value sites that fall outside #169's scope** — one because it is not a parsing bug at all, one because it is YAML. Neither overlaps with #169 or #171, so this can be worked in parallel with both.

Both were verified against `main` while triaging the audit.

### Site A — an unescaped title changes the OpenAlex query

`defendable-science/defendable_science/literature/acquire.py:713-715`:

```python
    page = client.get_json(
        f"{OPENALEX}/works",
        {"filter": f"title.search:{entry.title}", "per-page": str(SEARCH_LIMIT)},
    )
```

`entry.title` comes from the human's `references.json`. OpenAlex's `filter` grammar reserves `,` for AND and `|` for OR, so a title containing either — *"Deep Learning, Revisited"*, *"A|B testing"* — is parsed as two filter terms. The request succeeds. It just answers **a different question**, returning different siblings or none.

The consequence is not an error: the paper is routed to the `manual[]` worklist as though rung 4 had genuinely found nothing, when in fact rung 4 was never asked the intended question. This is precisely the "legitimately empty vs. we asked the wrong thing" distinction the module is otherwise scrupulous about — `_exhausted` (`acquire.py:1758`) exists to make exactly that distinction elsewhere.

No test covers it. `grep -rn "title.search" tests/` returns only a prose comment at `tests/test_acquire.py:80`, not an escaping test.

### Site B — `str()` used where validation is meant

`defendable-science/defendable_science/dataset/manifest.py:221-231` (`_decode_file_ref`):

```python
    try:
        path = str(raw["path"])
        sha256 = str(raw["sha256"])
    except KeyError as exc:
        raise ManifestError(f"{where}: file missing required key {exc}") from exc
```

The `isinstance(raw, dict)` guard above it is good, and a missing key raises properly. But `str()` on a *present* key of the wrong type never fails: a manifest written `path: [a, b]` silently becomes the string `"['a', 'b']"`. The same pattern is at `manifest.py:240` (`kind=str(raw.get("kind", ""))`) and `manifest.py:268` (`entry_id = str(raw["id"])`).

The later `validate()` pass catches a mangled checksum, because `SHA256_RE` (`manifest.py:24`) is applied at `manifest.py:393` — but nothing applies a comparable check to `path` or `id`. So the mangled path survives to `retrieval._resolve_file`, where `dataset/retrieval.py:96` does `Path(ref.path)` on it and looks for a file named `['a', 'b']`. The user sees a fixity or not-found failure pointing at the retrieval layer, several steps from the actual defect in their manifest.

## Goal

Neither boundary can produce a plausible wrong value: a reserved character in a title cannot silently change the OpenAlex query, and a wrong-typed scalar in `datasets.yml` is rejected at decode time with a message naming the entry and the field — not coerced into a string that fails later somewhere else.

## Where

- `defendable-science/defendable_science/literature/acquire.py:713-715` — site A. Check whether any other `filter:` expression in the module interpolates human-authored text; fix the class, not just this call.
- `defendable-science/defendable_science/dataset/manifest.py:221-231,236-242,262-269` — site B, the three `str()` coercion points.
- `defendable-science/defendable_science/dataset/manifest.py:24,393` — the existing `SHA256_RE` check, the model for what a real field check looks like here.
- `defendable-science/defendable_science/dataset/retrieval.py:96` — the downstream consumer that currently receives the mangled value; no change expected, cited so the blast radius is clear.
- `defendable-science/tests/test_acquire.py`, `defendable-science/tests/test_manifest.py` — regression tests.

## Proposed approach

**Site A.** Strip the reserved delimiters before interpolating, with a comment recording why — `title.search` is a full-text match, so dropping `,` and `|` narrows nothing that matters, and the three-way match gate re-checks the title against the real record afterwards (ADR-0037):

```python
#: OpenAlex's `filter` grammar reserves `,` (AND) and `|` (OR); a title carrying
#: either is parsed as two filter terms, silently answering a different query.
#: `title.search` is full-text, so replacing them narrows nothing — and the match
#: gate re-checks the title on the real record anyway.
_FILTER_RESERVED = re.compile(r"[,|]")

    search = _FILTER_RESERVED.sub(" ", entry.title)
```

Prefer this to URL-encoding: the characters are meaningful to OpenAlex's *filter grammar*, not to HTTP, so percent-encoding does not neutralise them.

**Site B.** Replace each `str(...)` with a checked accessor that raises `ManifestError` naming the entry, the field and the offending type — matching the message quality already set by `_int_or_error` (used at `manifest.py:231`) and by `manifest.py:264-267`. A small `_str_or_error(value, where, field)` helper alongside the existing `_int_or_error` and `_opt_str` is the least invasive fit; a `datasets.yml` Pydantic model is the alternative, but that is #171's YAML territory and should not be started here.

Accepting a plain `int` or `float` where a string is expected is a judgement call — `id: 2024` is a plausible thing for an author to write and arguably should be coerced. Decide it deliberately and state it in the docstring; the unambiguous requirement is that a **list or mapping** must never become a string.

## Out of scope

- `literature/graph.py:348` and `:110` — the other two silent-wrong-value sites, already covered by #169 defects 3 and 4. Do not touch `graph.py` here; #169 is in flight.
- The `dict[str, Any]` threading of the OpenAlex work record through fourteen `acquire.py` functions (`.codebase_audit/be-tech-debt.md` BE-9). Real debt, much larger, and it needs its own issue.
- Adding Pydantic to `datasets.yml`. That is the YAML boundary work in #171.

## Acceptance criteria

- [ ] A registry entry titled `Deep Learning, Revisited` produces an OpenAlex request whose `filter` contains exactly one `title.search:` term — regression test asserting the request the mocked transport actually received
- [ ] The same for a title containing `|`
- [ ] Any other human-authored text interpolated into a `filter:` expression in `acquire.py` is either escaped the same way or documented as safe, with the audit recorded in the PR
- [ ] `datasets.yml` with `path: [a, b]` raises `ManifestError` naming the entry and the field, rather than yielding the string `"['a', 'b']"` — regression test asserting the old symptom is gone
- [ ] Same for `sha256`, `retrieval.kind`, and a list- or mapping-valued `id`
- [ ] The error messages name the entry, the field and the offending type, matching the quality of `manifest.py:264-267`
- [ ] `dataset/retrieval.py` is unchanged — the fix is at the decode boundary, not downstream
- [ ] `literature/graph.py` is unchanged (owned by #169)
- [ ] `uv run pytest -q` passes with the 100% statement + branch gate intact; `uv run mypy` passes strict

## References

- #172 — the audit PR; `.codebase_audit/be-tech-debt.md` BE-4 and BE-9 are the source findings
- #169 — the JSON parsing-boundary work, which covers the two `graph.py` sites; in flight, do not collide
- #171 — the YAML boundary work
- `decisions/0037-literature-asset-acquisition.md` — the rung ladder and the three-way match gate that re-checks the title downstream
- `decisions/0009-dataset-thin-manifest.md` — the `datasets.yml` shape
- `CLAUDE.md` § Failure honesty — the rule both sites violate
- `defendable-science/defendable_science/literature/acquire.py:1758` (`_exhausted`) — the module's own precedent for distinguishing "legitimately empty" from "we could not ask properly"

davorrunje:
## What

Fixes two silent-wrong-value boundaries found by the codebase audit (#172,
BE-4/BE-9) that fell outside #169's JSON-parsing-boundary work: an unescaped
title corrupting an OpenAlex filter query, and `str()` coercion in the
manifest decoder silently stringifying lists/mappings.

## Details

- **Site A** (`literature/acquire.py`, `sibling_candidates`): OpenAlex's
  `filter` grammar reserves `,` (AND) and `|` (OR). A registry title
  containing either (e.g. "Deep Learning, Revisited") silently spliced a
  second filter clause into the rung-4 sibling search instead of erroring —
  the entry then landed on the `manual[]` worklist as if the search
  genuinely found nothing. Fixed with a `_FILTER_RESERVED` regex that
  replaces both characters with a space before interpolation (chosen over
  URL-encoding: the characters are meaningful to OpenAlex's filter grammar,
  not to HTTP). `title.search` is full-text, so this narrows nothing that
  matters, and the three-way match gate still re-verifies the real record
  afterwards (ADR-0037).
  - Audited every other `filter:`-interpolation site in `acquire.py`
    (`grep -n filter`) — the sibling-search call is the only one that builds
    an OpenAlex `filter` expression from human-authored text; no other site
    needed the fix.
- **Site B** (`dataset/manifest.py`): `_decode_file_ref`, `_decode_retrieval`,
  and `_decode_entry` used `str(...)` on a *present* key of the wrong type.
  A `KeyError` on a missing key was handled, but `path: [a, b]` in YAML
  silently became the string `"['a', 'b']"`, failing confusingly and far
  away in `dataset/retrieval.py`'s `Path(ref.path)`. Added `_str_or_error`
  (mirroring `_int_or_error`'s style/messages) and used it for `path`,
  `sha256`, `retrieval.kind`, and `id`. It accepts a real `str` or a bare
  `int`/`float` (coercing — `id: 2024` is a realistic thing to write in
  YAML) but raises `ManifestError` naming the entry, field, and offending
  type for a `list`/`dict`.
- Deliberately unchanged: `literature/graph.py` (owned by #169, already
  merged) and `dataset/retrieval.py` (downstream consumer — the fix is at
  the decode boundary only).

Closes #173

### DIFF
--- a/defendable-science/defendable_science/dataset/manifest.py
+++ b/defendable-science/defendable_science/dataset/manifest.py
@@ -268,8 +268,17 @@ def _decode_retrieval(raw: Any, where: str) -> Retrieval:
     """Decode a ``retrieval`` block."""
     if not isinstance(raw, dict):
         raise ManifestError(f"{where}: 'retrieval' must be a mapping")
+    kind_raw = raw.get("kind")
     return Retrieval(
-        kind=_str_or_error(raw.get("kind", ""), where, "retrieval.kind"),
+        # An explicit `kind: null` must resolve the same as an absent key —
+        # both mean "unset", and _decode_entry already treats a null
+        # `retrieval:` block itself as absent. Routing null through
+        # `_str_or_error` directly would abort the whole load() for a field
+        # that has a defined default, instead of the one clean validate()
+        # error the missing-key case produces.
+        kind=_str_or_error(
+            "" if kind_raw is None else kind_raw, where, "retrieval.kind"
+        ),
         url=_opt_str(raw.get("url")),
     )
 
--- a/defendable-science/defendable_science/dataset/manifest.py
+++ b/defendable-science/defendable_science/dataset/manifest.py
@@ -221,15 +221,44 @@ def _int_or_error(value: Any, where: str, fieldname: str) -> int | None:
         ) from exc
 
 
+def _str_or_error(value: Any, where: str, fieldname: str) -> str:
+    """Coerce a required field to ``str``, raising :class:`ManifestError` on shape.
+
+    Accepts a real ``str`` unchanged, and coerces a bare ``int``/``float`` (a
+    human writing ``id: 2024`` in YAML is realistic and unambiguous). A
+    ``list`` or ``dict`` is never silently stringified — the old
+    ``str(value)`` behavior turned a mistyped ``path: [a, b]`` into the
+    plausible-looking but wrong string ``"['a', 'b']"``, which then failed
+    far downstream instead of here. ``bool`` is rejected too: it is a
+    subclass of ``int`` but ``str(True)`` is not a value anyone meant to
+    write in this field.
+
+    :param value: The raw value to coerce.
+    :param where: Location prefix for the error message.
+    :param fieldname: The field name to name in the error message.
+    :returns: The coerced string.
+    :raises ManifestError: If `value` is not a string or scalar number.
+    """
+    if isinstance(value, str):
+        return value
+    if isinstance(value, (int, float)) and not isinstance(value, bool):
+        return str(value)
+    raise ManifestError(
+        f"{where}: '{fieldname}' must be a string, got {type(value).__name__}: {value!r}"
+    )
+
+
 def _decode_file_ref(raw: Any, where: str) -> FileRef:
     """Decode one ``files[]`` element, raising :class:`ManifestError` on shape."""
     if not isinstance(raw, dict):
         raise ManifestError(f"{where}: each file must be a mapping")
     try:
-        path = str(raw["path"])
-        sha256 = str(raw["sha256"])
+        path_raw = raw["path"]
+        sha256_raw = raw["sha256"]
     except KeyError as exc:
         raise ManifestError(f"{where}: file missing required key {exc}") from exc
+    path = _str_or_error(path_raw, where, "path")
+    sha256 = _str_or_error(sha256_raw, where, "sha256")
     return FileRef(
         path=path, sha256=sha256, size=_int_or_error(raw.get("size"), where, "size")
     )
@@ -240,7 +269,7 @@ def _decode_retrieval(raw: Any, where: str) -> Retrieval:
     if not isinstance(raw, dict):
         raise ManifestError(f"{where}: 'retrieval' must be a mapping")
     return Retrieval(
-        kind=str(raw.get("kind", "")),
+        kind=_str_or_error(raw.get("kind", ""), where, "retrieval.kind"),
         url=_opt_str(raw.get("url")),
     )
 
@@ -268,7 +297,7 @@ def _decode_entry(raw: Any, index: int) -> DatasetEntry:
         raise ManifestError(f"{where}: entry must be a mapping")
     if "id" not in raw:
         raise ManifestError(f"{where}: entry missing required key 'id'")
-    entry_id = str(raw["id"])
+    entry_id = _str_or_error(raw["id"], where, "id")
     where = f"entry '{entry_id}'"
 
     retrieval = None
--- a/defendable-science/defendable_science/literature/acquire.py
+++ b/defendable-science/defendable_science/literature/acquire.py
@@ -141,6 +141,15 @@ QUARANTINE_YEAR_WINDOW = 5
 _PUNCT = re.compile(r"[^\w\s]", flags=re.UNICODE)
 _SPACE = re.compile(r"\s+")
 
+#: OpenAlex's ``filter`` grammar reserves ``,`` (AND) and ``|`` (OR) between
+#: terms. A human-authored title containing either would silently splice a
+#: second filter clause into the query instead of erroring, so both are
+#: replaced with a space before interpolation. ``title.search`` is full-text,
+#: so dropping these characters narrows nothing that matters, and the
+#: three-way match gate re-checks the title against the real record
+#: afterwards (ADR-0037) — a query-time approximation is safe here.
+_FILTER_RESERVED = re.compile(r"[,|]")
+
 
 @dataclass
 class Candidate:
@@ -726,9 +735,10 @@ def sibling_candidates(
     if entry.title is None:
         return []
     anchor = _short_id(work.get("id"))
+    search = _FILTER_RESERVED.sub(" ", entry.title)
     page = client.get_json(
         f"{OPENALEX}/works",
-        {"filter": f"title.search:{entry.title}", "per-page": str(SEARCH_LIMIT)},
+        {"filter": f"title.search:{search}", "per-page": str(SEARCH_LIMIT)},
     )
     if not isinstance(page, dict):
         return []
--- a/defendable-science/defendable_science/literature/acquire.py
+++ b/defendable-science/defendable_science/literature/acquire.py
@@ -150,6 +150,14 @@ _SPACE = re.compile(r"\s+")
 #: afterwards (ADR-0037) — a query-time approximation is safe here.
 _FILTER_RESERVED = re.compile(r"[,|]")
 
+#: OpenAlex additionally reserves a *leading* ``!`` on a filter value as
+#: negation — ``title.search:!Kung`` means "does not contain Kung", not
+#: "contains !Kung". Only the leading position is reserved (a mid-title
+#: ``!``, e.g. "Kung! Fu", is just full-text content), so only a leading run
+#: of ``!`` is stripped rather than every ``!`` in the title — narrower than
+#: the comma/pipe case, where the character is reserved everywhere.
+_LEADING_BANG = re.compile(r"^!+\s*")
+
 
 @dataclass
 class Candidate:
@@ -735,7 +743,13 @@ def sibling_candidates(
     if entry.title is None:
         return []
     anchor = _short_id(work.get("id"))
-    search = _FILTER_RESERVED.sub(" ", entry.title)
+    search = _FILTER_RESERVED.sub(" ", entry.title).lstrip()
+    search = _LEADING_BANG.sub("", search).strip()
+    if not search:
+        # A title made entirely of reserved characters (",", "|", "!") would
+        # otherwise collapse to an empty or all-whitespace search, sending an
+        # unbounded `title.search:` query instead of a targeted one.
+        return []
     page = client.get_json(
         f"{OPENALEX}/works",
         {"filter": f"title.search:{search}", "per-page": str(SEARCH_LIMIT)},
--- a/defendable-science/defendable_science/literature/acquire.py
+++ b/defendable-science/defendable_science/literature/acquire.py
@@ -153,10 +153,16 @@ _FILTER_RESERVED = re.compile(r"[,|]")
 #: OpenAlex additionally reserves a *leading* ``!`` on a filter value as
 #: negation — ``title.search:!Kung`` means "does not contain Kung", not
 #: "contains !Kung". Only the leading position is reserved (a mid-title
-#: ``!``, e.g. "Kung! Fu", is just full-text content), so only a leading run
-#: of ``!`` is stripped rather than every ``!`` in the title — narrower than
-#: the comma/pipe case, where the character is reserved everywhere.
-_LEADING_BANG = re.compile(r"^!+\s*")
+#: ``!``, e.g. "Kung! Fu", is just full-text content), so only the leading
+#: run is stripped rather than every ``!`` in the title — narrower than the
+#: comma/pipe case, where the character is reserved everywhere.
+#:
+#: The class spans whitespace as well as ``!`` because :data:`_FILTER_RESERVED`
+#: runs *first* and substitutes a space for every ',' and '|'. That can
+#: manufacture a gap inside what was a single leading run — "!,!Kung" becomes
+#: "! !Kung" — and a pattern anchored to one unbroken run would leave the
+#: second ``!`` in place, inverting the query after all.
+_LEADING_BANG = re.compile(r"^[!\s]+")
 
 
 @dataclass
@@ -743,8 +749,7 @@ def sibling_candidates(
     if entry.title is None:
         return []
     anchor = _short_id(work.get("id"))
-    search = _FILTER_RESERVED.sub(" ", entry.title).lstrip()
-    search = _LEADING_BANG.sub("", search).strip()
+    search = _LEADING_BANG.sub("", _FILTER_RESERVED.sub(" ", entry.title)).strip()
     if not search:
         # A title made entirely of reserved characters (",", "|", "!") would
         # otherwise collapse to an empty or all-whitespace search, sending an
--- a/defendable-science/tests/test_acquire.py
+++ b/defendable-science/tests/test_acquire.py
@@ -710,6 +710,13 @@ def test_rung_4_comma_in_title_does_not_split_the_filter_query() -> None:
     An unescaped comma in the title would silently turn one
     ``title.search:`` clause into two ANDed clauses instead of raising, so
     the request actually sent must carry exactly one ``title.search:`` term.
+
+    The load-bearing check is the absence of ',' anywhere in the filter
+    value: counting terms with the ``title.search:`` prefix (as the first
+    assertion below does) is *not* discriminating on its own — with the sub
+    reverted, ``"title.search:Deep Learning, Revisited".split(",")`` still
+    yields exactly one element that starts with that prefix, so that
+    assertion alone would pass even with the bug present.
     """
     client = FakeClient({"/works": {"results": []}})
     entry = _entry(title="Deep Learning, Revisited")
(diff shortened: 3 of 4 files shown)


davorrunje:
## Code review — the two fixes are correct; findings are adjacent boundaries plus one regression

Reviewed at `590ccb6`. Gates from `defendable-science/`: **1804 passed, 28 skipped**, coverage **100.00%** (5232 stmts / 1544 branches, 0 missed), `ruff check` clean, `ruff format --check` clean, `mypy` clean.

Both fixes are genuine, and I mutation-checked them rather than taking the tests on trust:

- Reverting the `_FILTER_RESERVED.sub` → both new acquire tests fail.
- Collapsing `_str_or_error` back to `return str(value)` → 5 of the 6 new manifest tests fail.

No test pins third-party message text; every `pytest.raises(match=...)` targets our own `ManifestError` strings. Working tree left clean.

---

### 1. Regression — an explicit `kind: null` now aborts the whole file

`defendable-science/defendable_science/dataset/manifest.py`, the `kind=_str_or_error(...)` line.

```yaml
retrieval: {kind: null, url: http://a}
# this branch: ManifestError: 'retrieval.kind' must be a string, got NoneType: None  -> load() aborts
# main:        kind == ""  -> one clean validate() error, rest of the report intact

retrieval: {url: http://a}
# both:        kind == ""  -> one clean validate() error
```

On `main` the null and absent cases behaved identically. This branch splits them, which cuts against the treatment established in the three commits immediately preceding it — `48f7121` "stop treating an explicit JSON null as malformed input", `9e90b3a` "accept an explicit `distribution: null`", `cc2cada` "let a null contexts/intents keep the rest of its edge" — and against `_decode_entry`, which itself treats `retrieval: null` as absent.

It is a *loud* failure, so not a failure-honesty violation. But it costs the reader the rest of the validation report for a field that has a defined default. Suggest `raw.get("kind") if raw.get("kind") is not None else ""`, or routing null through the same absent-key path.

**This one is in scope for this PR** — it is a behaviour change introduced by these lines.

### 2. Targeting note — the two hardened fields were the two already defended

Worth recording, because it affects where the remaining risk sits:

- `retrieval.kind` was already caught loudly by the `RETRIEVAL_KINDS` membership check in `_validate_conditional`.
- `sha256` was already caught by `SHA256_RE` in `validate`.

The undefended fields are `path` (only `if not ref.path`) and `url` / `source` (no validation at all). Of those, only `path` got fixed here. Not a defect in this PR — the fixes are right — but the sweep is incomplete in a way the PR body's "audit" framing may obscure.

Filed as follow-ups rather than widening this PR:

- **#188** — `entry_from_croissant` still stringifies a JSON-LD node reference into a plausible-but-wrong `path`. `{"contentUrl": {"@id": "https://x/f.csv"}}` → `path == "{'@id': 'https://x/f.csv'}"`, truthy so the missing-field guard passes, survives `validate()`, reaches `Path(ref.path)`. This is the one genuinely third-party input path in the module, and the `CroissantDocument` docstring already invokes ADR-0043 point 4 for `name` one screen above.
- **#189** — `retrieval.url` is still `_opt_str`, one line below the field this PR hardened. A list `url` becomes `"['http://a', 'http://b']"`, `validate()` reports nothing, and `retrieval.py` hands that string to the Tier-B fetcher. Same for `source`, `pid`, `datasheet`, `version`, `tier`, `access`, and the `Citation` fields.
- **#190** — `arxiv_candidates` has the identical splice bug this PR fixed for OpenAlex: `f'ti:"{entry.title}"'`, so a title containing `"` terminates the phrase early. The audit grepped for `filter`; this is a `search_query` site. Failure mode is the same as the one fixed here — empty feed, entry lands on `manual[]` as though the rung legitimately found nothing.

### 3. The reserved-character class looks incomplete — in scope, since this PR owns the regex

Two gaps in `_FILTER_RESERVED`:

- OpenAlex reserves a **leading `!`** in a filter *value* for negation. A title beginning with `!` (e.g. `!Kung …`) yields `title.search:!Kung …` and **inverts** the search — silently returning an empty rung 4 rather than an error.
- A title consisting only of `,` / `|` collapses to whitespace and sends `filter=title.search: ` — an unbounded query.

Cheap fix while you are already in this regex: extend the class, and return `[]` when the sanitised search is empty.

### Nits

- `defendable-science/tests/test_acquire.py`, the comma test — its first assertion is non-discriminating for its own input. On `main`, `"title.search:Deep Learning, Revisited".split(",")` yields exactly one `title.search:`-prefixed term, so `len(title_search_terms) == 1` passes **with the bug present**. Only the final `"," not in …` line catches the regression (verified). Harmless, but the assertion reads as though it were the load-bearing one.
- `defendable-science/defendable_science/dataset/manifest.py`, the `not isinstance(value, bool)` guard — untested. Deleting it leaves all 1804 tests passing. The 100% branch gate cannot catch this because both conditions share a line. A one-line `id: true` case would pin it; float acceptance (`id: 1.5` → `"1.5"`) is likewise unpinned.

davorrunje:
Addressed at 0b40d6f.

**#1 (regression, required).** Reproduced: `retrieval: {kind: null, url: http://a}` raised `ManifestError` and aborted `load()`, while the absent-key case produced the single `retrieval.kind: must be one of [...]` validate() error. Fixed in `_decode_retrieval` — `kind_raw = raw.get("kind")`, then `"" if kind_raw is None else kind_raw` into `_str_or_error`, so null and absent now both resolve to `kind == ""`. Added `test_decode_retrieval_null_kind_matches_the_absent_key_case`, which asserts the null and absent manifests produce identical `validate()` error lists.

**#3 (regex, required).** Confirmed both gaps. Added `_LEADING_BANG = re.compile(r"^!+\s*")` and applied it after the existing comma/pipe substitution, stripping only a *leading* run of `!` (mid-title `!` is ordinary full-text content under `title.search`, so only the reserved leading position is touched — documented in the new comment). When the sanitised search is empty, `sibling_candidates` now returns `[]` before calling the client, instead of sending `filter=title.search:` unbounded. Added `test_rung_4_leading_bang_in_title_does_not_invert_the_search` and `test_rung_4_all_reserved_character_title_short_circuits_without_a_request` (the latter asserts `client.calls == []`).

**Nit 1.** Left the `len(title_search_terms) == 1` assertion in place as a sanity check but added a docstring stating plainly that it is not discriminating on its own (verified: it passes with the sub reverted) and that `"," not in params["filter"]` is the load-bearing check.

**Nit 2.** Added `test_decode_entry_id_bool_raises_naming_the_field` (`id: true` → `ManifestError` naming `'id'`/`bool`, pins the `not isinstance(value, bool)` guard) and `test_decode_entry_id_accepts_a_bare_float` (`id: 1.5` → `"1.5"`).

Full gate from `defendable-science/`: 1809 passed, 28 skipped, coverage 100.00% (5237 stmts / 1546 branches, 0 missed), `ruff check` clean, `ruff format --check` clean, `mypy` clean.

#188/#189/#190 acknowledged as out of scope and untouched.

davorrunje:
## Round-two review — all four prior findings closed; one residual of the same bug, worth landing before merge

Reviewed at `0b40d6f`. Gates from `defendable-science/`: **1809 passed, 28 skipped**, coverage **100.00%** (5237 stmts / 1546 branches, 0 missed), `ruff check` clean, `ruff format --check` clean, `mypy` clean. Diff vs `origin/main` (`9534127`, post-#185): 4 files, +237/−5.

### Prior findings

| # | Finding | Disposition |
|---|---|---|
| 1 | `kind: null` aborted `load()` | **RESOLVED** — `{kind: null, url: …}`, `{url: …}` and `{kind: '', …}` all load cleanly and yield the *identical* 8-error report ending in `retrieval.kind: must be one of [...]`. Mutating `"" if kind_raw is None else kind_raw` → `kind_raw` fails the new test. |
| 1b | Same null question for every `_str_or_error` field | **RESOLVED, no regression** — executed `path: null`, `sha256: null`, `id: null` against their absent counterparts; all six abort, so null and absent stay behaviourally equivalent for the three required keys. Worth noting this closed a silent-wrong-value the first round missed: pre-PR, `path: null` quietly became the string `"None"`. |
| 2 | `_FILTER_RESERVED` incomplete | **PARTIALLY RESOLVED** — both gaps addressed and mutation-verified (leading `!` stripped; all-reserved title short-circuits with zero HTTP calls), but the `!` regex is under-strict — see below. |
| 3 | Non-discriminating first assertion in the comma test | **RESOLVED** — assertion kept, docstring now states outright that it is not discriminating and names the `"," not in` line as load-bearing. Good call; deleting it would have lost the readable example. |
| 4 | `not isinstance(value, bool)` untested; float unpinned | **RESOLVED** — both mutation-checked: dropping the bool guard fails `test_decode_entry_id_bool_raises_naming_the_field`; narrowing `(int, float)` → `int` fails `test_decode_entry_id_accepts_a_bare_float`. |

Also confirmed: **#188 / #189 / #190 are untouched** — no partial fixes, so those issue texts remain accurate. Verified `retrieval: {kind: http, url: [a, b]}` still loads with `url == "['a', 'b']"` via `_opt_str` (#189), and `arxiv_candidates` still carries `query = f'ti:"{entry.title}"'` (#190).

No test in the diff pins third-party message text — every `pytest.raises(match=...)` targets a `ManifestError` string this package owns.

---

### New finding — `_LEADING_BANG` strips only one bang-run, so the inverted-search bug survives

`defendable-science/defendable_science/literature/acquire.py:159` (`_LEADING_BANG = re.compile(r"^!+\s*")`), applied at `:747`.

The regex matches one `!`-run plus its trailing whitespace and cannot repeat. But `_FILTER_RESERVED.sub(" ", …)` runs **first** and converts `,` / `|` into spaces — so it manufactures the interleaved shape the regex cannot clear:

```
title = "!,!Kung Fu"
  _FILTER_RESERVED.sub(" ", …)  ->  "! !Kung Fu"
  .lstrip()                      ->  unchanged (starts with '!')
  _LEADING_BANG.sub("", …)       ->  consumes "! "  ->  "!Kung Fu"
  request: filter=title.search:!Kung Fu
```

OpenAlex reads the leading `!` as **negation** and returns works that do *not* match the title. Rung 4 comes back empty, indistinguishable from a genuine miss — precisely the defect this fixup set out to close. Same for `"! !Kung"` and `"!|!Kung"`.

Fix is one character class:

```python
_LEADING_BANG = re.compile(r"^[!\s]+")
```

applied after the reserved-character substitution — which also makes the `.lstrip()` at `:746` redundant and collapses `:746-747` into a single pass. Suggested test: `title="!,!Kung Fu"` asserting `params["filter"] == "title.search:Kung Fu"`.

The constant's docstring ("only a leading run of `!` is stripped") is accurate to the code, but the code is under-strict for what it intends — worth updating alongside.

**I'd land this before merge** rather than as a follow-up: it is the same defect class the PR exists to close, and it is a one-character change plus one test.

### Deferred to issues

Neither blocks:

- **#193** — the sanitised-empty short-circuit returns a bare `[]`, so a *skipped* rung is indistinguishable from a legitimate miss. Flagged as informational, not a new violation: it is exactly consistent with the other bail-outs in the same function (`entry.title is None`, `not isinstance(page, dict)`), and the `list[Candidate]` return type offers no degradation channel. Filed as a design issue covering all three bail-outs rather than patched here.
- **#194** — `path: null` reports `'path' must be a string, got NoneType: None` where the actionable message is "you left it unset". Behaviour is identical to the absent-key case; wording only.

### Recommendation

**Approve once the `_LEADING_BANG` one-liner lands.** All four prior findings are genuinely closed and mutation-verified, the null-equals-absent contract is intact, and there is no encroachment on the deferred issues.

davorrunje:
## Pushed `0282c5e` — the `_LEADING_BANG` residual is fixed

Written test-first. The new test fails on `0b40d6f` for exactly the right reason:

```
E       AssertionError: assert 'title.search:!Kung Fu' == 'title.search:Kung Fu'
```

— i.e. the inverted query really was going out over the wire, not just theoretically.

**The fix** is the one-character-class change: `^!+\s*` → `^[!\s]+`. Since the class now spans leading whitespace, the preceding `.lstrip()` became redundant and the two passes collapse into one:

```python
search = _LEADING_BANG.sub("", _FILTER_RESERVED.sub(" ", entry.title)).strip()
```

The constant's docstring gains a paragraph explaining *why* the class spans whitespace — that `_FILTER_RESERVED` runs first and can manufacture the gap — so the next reader does not "simplify" it back to `^!+`.

**Sanitisation across the cases:**

| title | → search |
|---|---|
| `!,!Kung Fu` | `Kung Fu` |
| `! !Kung` | `Kung` |
| `!|!Kung` | `Kung` |
| `!Kung-Fu Panda` | `Kung-Fu Panda` |
| `  !Kung` | `Kung` |
| `Kung! Fu` | `Kung! Fu` ← mid-title `!` still preserved |
| `!,\|` | `` ← still short-circuits, no request |

The documented "narrower than the comma/pipe case" intent holds: only the leading run goes.

**Gates:** 1810 passed, 28 skipped (+1 from the new test), coverage 100.00% (5236 stmts / 1546 branches, 0 missed). `ruff check`, `ruff format --check`, `mypy` all clean. `git merge-tree` against current `main` (`1d639cf`, post-#187) reports a clean merge.

Nothing else touched — #188 / #189 / #190 / #193 / #194 remain out of scope and their issue texts stay accurate.

From my side this now looks mergeable.

PR #187, opened 2026-08-29
fix(literature): require WorksPage's results/meta (closes #176)

### LINKED ISSUE #176: literature: make WorksPage's results and meta required so a malformed OpenAlex page cannot read as empty

davorrunje:
## What

`WorksPage.results`/`.meta` and `_PageMeta.next_cursor` defaulted, so a
structurally malformed OpenAlex `/works` page still validated: a missing
`meta` mid-pagination silently truncated the citation frontier as if
complete, and an HTTP-200 OpenAlex error envelope silently read as "zero
citations". Both now raise a named `HttpError` naming the source URL and
the missing field, matching `cites()`'s own documented hard-fail contract.

## Details

- `results`/`meta` have no default at all — a missing key is now a
  validation failure, and `parse_obj` already turns that into the
  `HttpError` `cites()` expects, so no new error handling was needed there.
- `meta` deliberately stays `_PageMeta | None` rather than becoming fully
  non-nullable: OpenAlex sends an explicit `"meta": null` mid-pagination to
  mean "no further cursor" (a prior fix in #179 added a regression test for
  exactly this), and that must keep working — only a *missing* `meta` key
  is the defect being closed here.
- `_PageMeta.next_cursor` drops its default (required key) but keeps
  `str | None` (nullable value), since OpenAlex always sends the key but
  its value is genuinely `null` on the final page.
- Three new regression tests: missing `meta` raises, an HTTP-200 error
  envelope raises instead of returning `[]`, and a final page with
  `meta.next_cursor: null` still terminates pagination normally.
- The S2 models in the same module (best-effort, skip-and-mark boundary)
  are unchanged, per the issue's explicit scope.
- `pytest` (100% statement+branch coverage gate), `mypy --strict`,
  `ruff check`, and `ruff format --check` all pass.

Closes #176

### DIFF
--- a/defendable-science/defendable_science/literature/graph.py
+++ b/defendable-science/defendable_science/literature/graph.py
@@ -85,23 +85,35 @@ class OpenAlexWork(ExternalModel):
 
 
 class _PageMeta(ExternalModel):
-    next_cursor: str | None = None
+    """OpenAlex's per-page pagination metadata.
+
+    ``next_cursor`` has no default: OpenAlex always sends the key, so a
+    missing one is a malformed body, not "no further page" — but the *value*
+    stays ``str | None`` because ``null`` is exactly how OpenAlex spells "no
+    further page" on the last one. Required key, nullable value.
+    """
+
+    next_cursor: str | None
 
 
 class WorksPage(ExternalModel):
     """One cursor-paginated page of the OpenAlex ``/works`` endpoint.
 
-    ``meta`` tolerates an explicit ``null`` as well as a missing key —
-    unlike ``results``, a `default_factory` alone is not enough:
-    ``strict=True`` still rejects a present-but-``null`` value, and OpenAlex
-    sending ``"meta": null`` unambiguously means "no further cursor", not a
-    malformed page. Hard-failing there would discard every page already
-    collected for no benefit (see :func:`cites`'s truncated-frontier
-    argument, which rests on ``results`` staying strict, not on ``meta``).
+    ``results`` has no default: OpenAlex always sends it on a successful
+    response, so a page missing it — including an HTTP-200 error envelope
+    like ``{"error": ..., "message": ...}`` — is malformed, not empty.
+
+    ``meta`` likewise has no default, so a missing key is malformed too, but
+    its type stays ``_PageMeta | None``: OpenAlex sends an explicit
+    ``"meta": null`` mid-pagination to mean exactly "no further cursor", and
+    that is a legitimate final page, not a defect. Hard-failing a missing
+    ``meta`` key while still accepting a present ``null`` value is what keeps
+    :func:`cites` from reading a truncated frontier as complete (see its
+    docstring) without discarding a page OpenAlex genuinely ended.
     """
 
-    results: list[OpenAlexWork] = Field(default_factory=list)
-    meta: _PageMeta | None = None
+    results: list[OpenAlexWork]
+    meta: _PageMeta | None
 
 
 class _ExternalIdBundle(ExternalModel):
--- a/defendable-science/defendable_science/literature/graph.py
+++ b/defendable-science/defendable_science/literature/graph.py
@@ -103,13 +103,15 @@ class WorksPage(ExternalModel):
     response, so a page missing it — including an HTTP-200 error envelope
     like ``{"error": ..., "message": ...}`` — is malformed, not empty.
 
-    ``meta`` likewise has no default, so a missing key is malformed too, but
-    its type stays ``_PageMeta | None``: OpenAlex sends an explicit
-    ``"meta": null`` mid-pagination to mean exactly "no further cursor", and
-    that is a legitimate final page, not a defect. Hard-failing a missing
-    ``meta`` key while still accepting a present ``null`` value is what keeps
-    :func:`cites` from reading a truncated frontier as complete (see its
-    docstring) without discarding a page OpenAlex genuinely ended.
+    ``meta`` likewise has no default, so a missing key is malformed too. Its
+    type stays ``_PageMeta | None``, tolerating an explicit ``"meta": null``
+    as a defensive fallback rather than something the live API is known to
+    send: OpenAlex always returns a ``meta`` object, and the last page is
+    spelled ``meta.next_cursor: null`` (see :class:`_PageMeta`). Hard-failing
+    a missing ``meta`` key is what keeps :func:`cites` from reading a
+    truncated frontier as complete (see its docstring); whether tolerating an
+    explicit top-level ``null`` should itself become a hard failure is
+    tracked separately (defendable-science#191).
     """
 
     results: list[OpenAlexWork]
@@ -159,12 +161,10 @@ class S2CitationEdge(ExternalModel):
 class S2CitationsPage(ExternalModel):
     """One page of S2's ``/paper/{id}/citations`` response.
 
-    ``data`` has **no default** on purpose: unlike ``WorksPage.results``
-    (which defaults to ``[]`` because a missing/empty page is not a hard
-    error there either), a missing, ``null``, or non-list ``data`` here is a
-    malformed page, not a page with zero edges — the two must not collapse,
-    or a truncated response would read as "this work simply has no citation
-    edges."
+    ``data`` has **no default** on purpose, like ``WorksPage.results``: a
+    missing, ``null``, or non-list ``data`` here is a malformed page, not a
+    page with zero edges — the two must not collapse, or a truncated response
+    would read as "this work simply has no citation edges."
     """
 
     data: list[Any]
--- a/defendable-science/tests/test_literature.py
+++ b/defendable-science/tests/test_literature.py
@@ -1571,3 +1571,69 @@ def test_enrich_marks_is_influential_degraded_even_when_samples_survive() -> Non
     assert rec["intent"] == "bg"
     assert rec["is_influential"] is False  # the value is unreliable, not corrected
     assert rec["degraded"] == ["is_influential"]
+
+
+# --- boundary-validation regressions (defendable-science#176) --------------
+
+
+def test_cites_missing_meta_key_raises_instead_of_truncating() -> None:
+    """A page missing `meta` entirely must not read as "no further cursor".
+
+    `meta: _PageMeta | None = None` used to default a *missing* key to
+    `None`, silently ending pagination as if the frontier were complete.
+    `meta` now has no default, so a missing key is the malformed page it
+    actually is — and `cites` needs no new error handling to see it, because
+    `parse_obj` already turns the validation failure into a named `HttpError`.
+    """
+    client = _client(
+        {
+            "https://api.openalex.org/works": {
+                "results": [{"id": "https://openalex.org/W2"}]
+            }
+        }
+    )
+    with pytest.raises(
+        http.HttpError, match=r"https://api\.openalex\.org/works.*meta.*Field required"
+    ):
+        graph.cites("W1", client=client)
+
+
+def test_cites_error_envelope_raises_instead_of_reporting_zero_citations() -> None:
+    """An OpenAlex HTTP-200 error envelope must not read as "zero citations".
+
+    `results: list[OpenAlexWork] = Field(default_factory=list)` used to
+    default a *missing* `results` key to `[]`, so `{"error": ...,
+    "message": ...}` — OpenAlex's own error shape, delivered with a 200 —
+    silently became an empty, "legitimate" citation list instead of the
+    malformed page it is.
+    """
+    client = _client(
+        {
+            "https://api.openalex.org/works": {
+                "error": "Invalid query parameters",
+                "message": "cites filter value is not a valid OpenAlex ID",
+            }
+        }
+    )
+    with pytest.raises(
+        http.HttpError,
+        match=r"https://api\.openalex\.org/works.*results.*Field required",
+    ):
+        graph.cites("W1", client=client)
+
+
+def test_cites_null_next_cursor_still_terminates_pagination() -> None:
+    """A final page with `meta.next_cursor` explicitly `null` is not malformed.
+
+    `next_cursor` now has no default of its own, but its *value* legitimately
+    is `None` on the last page — required key, nullable value — so a
+    well-formed final page must keep terminating pagination normally rather
+    than becoming a stricter regression than the model it replaced.
+    """
+    page = {
+        "results": [{"id": "https://openalex.org/W2"}],
+        "meta": {"next_cursor": None},
+    }
+    client = _client({"https://api.openalex.org/works": page})
+    rows = graph.cites("W1", client=client)
+    assert [r["id"]["openalex"] for r in rows] == ["W2"]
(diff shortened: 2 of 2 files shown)


davorrunje:
## Code review — behaviour change is correct and matches the live API; four documentation/robustness findings

Reviewed at `5a61815`. Gates from `defendable-science/`: **1799 passed, 28 skipped**, coverage **100.00%** (statement+branch), `ruff check` clean, `ruff format --check` clean, `mypy` clean. Working tree left clean.

Because the whole change rests on empirical claims about OpenAlex's response shape, I checked them against the live API rather than the docstrings:

- **Non-final cursor page** — `meta` is always an object containing `next_cursor` (plus `count`, `page`, `x_query`, …). Requiring `results` / `meta` / `next_cursor` is safe.
- **Genuine zero-hit search** (`filter=cites:W9999999999999`) — returns `{"meta": {…, "next_cursor": null}, "results": []}`, parses cleanly to `results=[]`, and `cites()` returns `[]`. **The legitimately-empty case is not misclassified as a failure** — verified directly through `parse_obj(WorksPage, …)`.
- **Bad filter value** — OpenAlex answers **HTTP 400**, not a 200 error envelope.

The core change is right: `cites()` is the only consumer of `WorksPage`, its pagination loop still handles the `meta`-is-`None` branch, and the `max_results` early return is unaffected.

---

### 1. The `S2CitationsPage` docstring now states the opposite of what this PR just changed

It reads: *"`data` has **no default** on purpose: unlike `WorksPage.results` (which defaults to `[]` because a missing/empty page is not a hard error there either)…"*

As of this PR `WorksPage.results` has no default either, and a missing key **is** a hard error there. A future maintainer reading `S2CitationsPage` is told `WorksPage` deliberately tolerates a missing `results` — and may reintroduce exactly the `Field(default_factory=list)` that #176 exists to remove. Docstrings are this repo's rationale record, so a stale one here is a live regression risk.

Fix: drop the parenthetical, or invert it to "like `WorksPage.results`".

### 2. The `meta` escape hatch is justified by a claim the live API contradicts

The new `WorksPage.meta` docstring asserts that *"OpenAlex sends an explicit `"meta": null` mid-pagination to mean exactly 'no further cursor'"*.

Live OpenAlex never does this — it always sends a `meta` **object** and spells "last page" as `meta.next_cursor: null`. That is also what `_PageMeta`'s own docstring says two paragraphs earlier, so the two docstrings now contradict each other.

The consequence is the failure mode this PR's title claims to close. If anything — a proxy, a cache, a truncated body, a future API change — returns `{"results": [<200 works>], "meta": null}` on page 1 of a 1,245-citation frontier, then `page.meta` is `None`, `cursor` becomes `None`, and `cites()` returns 200 rows **reporting a truncated frontier as complete**. Silently. The existing green test `test_cites_null_meta_ends_pagination_without_discarding_the_page` pins precisely this behaviour.

This is **inherited from #179** (`9e90b3a` / `cc2cada`), not introduced here, so I have filed it as **#191** rather than asking you to widen this PR. But this PR re-derives and re-blesses the tolerance with a rationale the API contradicts — so at minimum, please correct the docstring text here even if the tightening lands in #191.

### 3. The new test's stated premise does not match the real API

In `defendable-science/tests/test_literature.py`, the new test docstring says OpenAlex's error shape is *"delivered with a 200"*. It is delivered with **HTTP 400** (verified), which `HttpClient.get_json` already turns into an `HttpError` before `WorksPage` is ever reached.

The *code* change is still correct — any 200 body missing `results` is malformed regardless of provenance — but the test's justification is fictional, which weakens it as documentation of why the strictness is needed. Suggest rewording to something like "any 200 body missing `results` — e.g. an error envelope surfaced through a proxy that rewrites the status".

### 4. Both new `pytest.raises` regexes pin Pydantic's own message text

`"Field required"` is Pydantic's `missing` error message, and `defendable-science/pyproject.toml` allows `pydantic>=2.13,<3`. The string has been stable across 2.x, but nothing in the pin guarantees it; a wording change in any allowed release turns these into false failures.

The stable contract is the error *type* (`missing`) and the field path, both of which `_explain` already emits — matching on the field path alone (e.g. `r"meta: .*\(got dict\)"`) would be version-proof. The same pattern exists elsewhere in the suite, so this is consistency-preserving rather than novel; flagging it because the identical issue was just raised on #185.

---

**Summary:** no correctness defect in the shipped behaviour change. Required `results` / `meta` / `next_cursor` matches the live OpenAlex contract, genuine zero-hit pages still read as legitimately empty, and malformed pages now raise a named `HttpError` carrying the field path. Findings 1–3 are documentation accuracy; the underlying robustness gap behind finding 2 is tracked in **#191**.

davorrunje:
Fixed in ca8fb2d.

1. `S2CitationsPage` docstring: dropped the stale parenthetical claiming `WorksPage.results` defaults to `[]`. Both fields now state they behave the same way (no default, missing key is a hard error).

2. `WorksPage.meta` docstring: removed the claim that live OpenAlex sends an explicit `"meta": null` mid-pagination. Now states what `_PageMeta` already documents — OpenAlex always returns a `meta` object, and the last page is spelled `meta.next_cursor: null` — and that tolerating an explicit top-level `null` is a defensive fallback, not observed API behavior. No code/behavior change; the field stays nullable. #191 is acknowledged as tracking whether that tolerance should become a hard failure — not implemented here.

3. `test_cites_error_envelope_raises_instead_of_reporting_zero_citations` docstring: reworded "delivered with a 200" (verified HTTP 400) to "a 200 body shaped like `{"error": ..., "message": ...}` — e.g. an error envelope surfaced through a proxy that rewrites the status." Test assertions unchanged.

4. Both `pytest.raises(match=...)` regexes in `test_cites_missing_meta_key_raises_instead_of_truncating` and `test_cites_error_envelope_raises_instead_of_reporting_zero_citations` now match on field path + received type (`r"meta: .*\(got dict\)"`, `r"results: .*\(got dict\)"`) instead of Pydantic's `"Field required"` message text, per the field-path-only idiom used elsewhere (`test_models.py`).

Verified against `_explain` in `core/models.py` and against a live parse of both malformed shapes before changing the regexes.

Gates from `defendable-science/`: 1799 passed, 28 skipped, coverage 100.00%, `ruff check` clean, `ruff format --check` clean, `mypy` clean.

davorrunje:
## Round-two review — approve; one residual docstring sentence

Reviewed at `ca8fb2d`. Gates from `defendable-science/`: **1799 passed, 28 skipped**, coverage **100.00%** (5222 stmts / 1540 branches, 0 missed; `graph.py` 290/108 at 100%), `ruff check` clean, `ruff format --check` clean, `mypy` clean. Diff vs `origin/main` (`9534127`, post-#185): 2 files, +95/−16. Branch is behind `main`, but `git merge-tree --write-tree origin/main HEAD` reports a clean merge — the only overlap is a disjoint region of `tests/test_literature.py`.

### Prior findings

| # | Finding | Disposition |
|---|---|---|
| 1 | Stale `S2CitationsPage` parenthetical | **RESOLVED** — now "``data`` has **no default** on purpose, like ``WorksPage.results``". Accurate: both are bare required fields, and under `strict=True` a missing/`null`/non-list value fails for each. No residual claim about a default. |
| 2 | False `"meta": null` rationale | **RESOLVED** — now calls the tolerance "a defensive fallback rather than something the live API is known to send" and states "OpenAlex always returns a ``meta`` object, and the last page is spelled ``meta.next_cursor: null``". Both live probes confirm this exactly, and the two docstrings now agree. The `defendable-science#191` pointer is correct (issue is open) and matches the repo's existing cross-ref style (`cli.py:65`, `:99`, `:152`). Tightening correctly left out of scope. |
| 3 | Fictional "delivered with a 200" | **PARTIALLY RESOLVED** — fixed in the test docstring, survives in the model docstring. See below. |
| 4 | `pytest.raises` pinning `"Field required"` | **RESOLVED** — both regexes now read `r"https://api\.openalex\.org/works.*<field>: .*\(got dict\)"`, pinning our own `_explain` format plus the field path. `grep -rn "Field required" tests/` → zero hits. This is the more robust convention, and I have generalised it to the ten pre-existing sites elsewhere in the suite as **#192**. |

### Behavioural claims re-verified independently at HEAD

Not taken from round one — re-probed against the live API and replayed through the code at `ca8fb2d`:

- `GET /works?filter=cites:W9999999999999&per-page=200&cursor=*` → **HTTP 200**, `{"meta": {"count": 0, …, "next_cursor": null, …}, "results": [], "group_by": []}`. Fed that real body through HEAD: `results=[]`, `meta.next_cursor=None`, `cites(...)` → `[]`. **Legitimately empty, not a failure — no regression.** This was the main risk in the change and it is clean.
- Real last page (`meta` present, `next_cursor: null`, non-empty `results`) → walk terminates cleanly and returns accumulated rows; a two-page walk yields `["W2","W3"]` correctly.
- Malformed pages all raise a named `HttpError` carrying the field path, never a raw `ValidationError`: `{"results":[]}` → `meta: Field required (got dict)`; `{"meta":{…}}` → `results: …`; `{"error":…}` → both; `{"results":[],"meta":{}}` → `meta.next_cursor: Field required (got dict)`.
- Consumer trace: `WorksPage` has exactly **one** consumer (`cites()`) and exactly **one** pagination loop. `_PageMeta` is used only by `WorksPage`. `resolve` and `refs` use different models. No other endpoint affected.
- The deferred gap reproduces as documented: `{"results":[],"meta":null}` → `[]` silently. That is #191, correctly scoped out.

`ca8fb2dd` introduces **no behavioural change** — docstrings plus the two regexes. Every other factual assertion in the new text was checked against the live API or the code and holds.

---

### New finding (low) — the model docstring still carries the claim the fixup corrected in the test

`defendable-science/defendable_science/literature/graph.py:103-105`:

> ``results`` has no default: OpenAlex always sends it on a successful response, so a page missing it — including an HTTP-200 error envelope like ``{"error": ..., "message": ...}`` — is malformed, not empty.

OpenAlex delivers that envelope with **HTTP 400**, not 200 — verified live (`filter=cites:NOTANID` → 400, `{"error":"Invalid query parameters error.", …}`), and `HttpClient` converts it to `HttpError` at `core/http.py:351` before `WorksPage` is ever reached.

The fixup rewrote the *test* docstring to the honest framing ("an error envelope surfaced through a proxy that rewrites the status") but left the model docstring — the canonical rationale record — carrying the unqualified version. Given that a confidently-worded false claim in exactly this docstring is what produced finding 2 in the first place, it is worth converging the wording. Suggested: mirror the test's phrasing, e.g. "— including any 200 body shaped like an error envelope, e.g. one surfaced through a proxy that rewrites the status —".

### Recommendation

**Approve.** The PR does what #176 asked, the honesty-critical zero-hit path is verified intact against a real OpenAlex response, and the deferred `meta: null` gap is correctly scoped out and tracked in an open #191. The finding above is a one-sentence docstring nit — worth a follow-up commit on this branch given the repo's "docstrings are the rationale record" posture, but it does not block.

